PR #4, opened 2017-09-16
Consul 0.9.3 and standardize to official Terraform module spec.

josh-padnick:
This PR updates Consul to the latest version at the time of this PR, v0.9.3. Beyond a simple version bump, the only meaningful change was updating the Consul configuration from using a `retry_join_ec2` directive to the now required (and cloud-agnostic) `retry_join`.

This also fixes a subtle issue where, on Amazon Linux, supervisord is correctly installed but does not immediately start. This issue is only relevant when users manually run the `install-consul` script followed by the `run-consul` script without rebooting first.


josh-padnick:
Tests are failing due to permission issue with HashiCorp org and CircleCI. Merging now but feedback very welcome on this!

PR #6, opened 2017-09-19
Fix READMEs and bug in example startup script

brikis98:
1. The latest version of Consul uses `-http-addr` and not `rpc-addr`. Update the example script to use it.
1. Fix some minor issues in the READMEs.


brikis98:
Simple fix. Merging now. Feedback welcome!

PR #9, opened 2017-09-25
ability to supply vpc_id

andrewmeissner:
Adds the ability to optionally deploy the consul agents to a specified VPC.  The default behavior is still maintained, in that if `vpc_id` is not supplied, the default VPC is used.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -152,7 +152,7 @@ data "template_file" "user_data_client" {
 
 data "aws_vpc" "default" {
   default = "${var.vpc_id == "" ? true : false}"
-  id = "${var.vpc_id == "" ? "" : var.vpc_id}"
+  id = "${var.vpc_id}"
 }
 
 data "aws_subnet_ids" "default" {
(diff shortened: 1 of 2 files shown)


andrewmeissner:
:(  Can someone help give me insight as to why the checks failed?

main.tf
```diff
@@ -151,7 +151,8 @@ data "template_file" "user_data_client" {
 # ---------------------------------------------------------------------------------------------------------------------
 
 data "aws_vpc" "default" {
-  default = true
+  default = "${var.vpc_id == "" ? true : false}"
+  id = "${var.vpc_id == "" ? "" : var.vpc_id}"
```
brikis98: Couldn't you just set the `id` param to `var.vpc_id` without the if-statement? It seems like it would have no effect.

Also, does the `aws_vpc` data source work correctly if you explicitly set `id` to an empty string?

andrewmeissner: @brikis98, you're absolutely right about the `id` param.  (facepalm)

I believe the data source does work correctly.  I ran terraform both explicitly specifying the `vpc_id` in the module as well as omitting the parameter.  In the former case the instances were associated with the specified vpc, and in the latter, the default vpc was used.

brikis98:
https://github.com/hashicorp/terraform-aws-consul/releases/tag/v0.0.3

andrewmeissner:
Thanks!

PR #11, opened 2017-09-27
Nit in install-dnsmasq

aaron0browne:
Replace `/etc/dnsmasq.d` literal with `DNS_MASQ_CONFIG_DIR` var.

### DIFF
--- a/modules/install-dnsmasq/install-dnsmasq
+++ b/modules/install-dnsmasq/install-dnsmasq
@@ -86,7 +86,7 @@ function install_dnsmasq {
     sudo yum update -y
     sudo yum install -y dnsmasq
     echo "prepend domain-name-servers $consul_ip;" | sudo tee -a "/etc/dhcp/dhclient.conf" > /dev/null
-    echo "conf-dir=/etc/dnsmasq.d" | sudo tee -a "/etc/dnsmasq.conf" > /dev/null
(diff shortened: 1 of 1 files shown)


modules/install-dnsmasq/install-dnsmasq
```diff
@@ -86,7 +86,7 @@ function install_dnsmasq {
     sudo yum update -y
     sudo yum install -y dnsmasq
     echo "prepend domain-name-servers $consul_ip;" | sudo tee -a "/etc/dhcp/dhclient.conf" > /dev/null
-    echo "conf-dir=/etc/dnsmasq.d" | sudo tee -a "/etc/dnsmasq.conf" > /dev/null
+    echo "conf-dir=$DNS_MASQ_CONFIG_DIR" | sudo tee -a "/etc/dnsmasq.conf" > /dev/null
```
brikis98: Instead of an somewhat hidden environment variable that no one will know to set, you could you add this as a command-line arg to the script (following the pattern of all the args already there) and pass it through to this function?

aaron0browne: This variable is already set on [line 14](https://github.com/aaron0browne/terraform-aws-consul/blob/0a5e90055b8c38d18a5543d73fefcaf11347752a/modules/install-dnsmasq/install-dnsmasq#L14).

brikis98: Ohh, haha, you're right. OK, looks like I missed that while trying to make the code more DRY :)

brikis98:
Thanks for the PR! Merging now.

PR #14, opened 2017-10-10
Support source security group id ingress access to cluster

sclausson:


sclausson:
Plan runs ok locally, so not sure why/where the automated tests are failing.

modules/consul-cluster/main.tf
```diff
@@ -95,6 +95,7 @@ resource "aws_security_group" "lc_security_group" {
 }
 
 resource "aws_security_group_rule" "allow_ssh_inbound" {
+  count       = "${length(var.allowed_inbound_cidr_blocks) >= 1 ? 1 : 0}"
```
brikis98: Should this be `allowed_ssh_cidr_blocks`?

modules/consul-cluster/variables.tf
```diff
@@ -24,6 +24,11 @@ variable "allowed_inbound_cidr_blocks" {
   type        = "list"
 }
 
+variable "allowed_inbound_security_group_ids" {
+  description = "A list of security group IDs that will be allowed to connect to Consul"
+  type        = "list"
```
brikis98: For backwards compatibility, should probably default this to an empty list.

modules/consul-security-group-rules/main.tf
```diff
@@ -92,3 +101,101 @@ resource "aws_security_group_rule" "allow_dns_udp_inbound" {
   security_group_id = "${var.security_group_id}"
 }
 
+resource "aws_security_group_rule" "allow_server_rpc_inbound_from_security_group_ids" {
+  count                    = "${length(var.allowed_inbound_security_group_ids)}"
```
brikis98: Is `allowed_inbound_security_group_ids` a new input variable for the consul-security-group-rules module? If so, don't you need to add it to its `vars.tf` and pass it in from the `consul-cluster` module?

sclausson: Good catch. I have made these changes.

modules/consul-cluster/variables.tf
sclausson: I've now set this to default = []

modules/consul-cluster/main.tf
sclausson: Good catch. I've corrected this.

sclausson:
I think this should be ok now.

modules/consul-security-group-rules/variables.tf
```diff
@@ -12,6 +12,11 @@ variable "allowed_inbound_cidr_blocks" {
   type        = "list"
 }
 
+variable "allowed_inbound_security_group_ids" {
+  description = "A list of security group IDs that will be allowed to connect to Consul"
+  type        = "list"
```
brikis98: This should probably default to an empty list for backwards compatibility

sclausson:
@brikis98 I've made one final requested change and added a default value to `allowed_inbound_security_group_ids` in `consul-security-group-rules/variables.tf` I think this is ready to merge.

brikis98:
Fantastic, thank you!

brikis98:
https://github.com/hashicorp/terraform-aws-consul/releases/tag/v0.0.4

PR #25, opened 2017-11-13
Update consul-cluster documentation

bizmate:
I have tried to work out a correct minimal consul cluster example based on the previous old documentation. I think this is an improvement but I am getting error 500 on the cluster instances

```
HTTP error code from Consul: 500 Internal Server Error

Error message from Consul: No cluster leader
```
I am unsure what is needed to get this fixed. I am opening the PR to both fix the documentation and be able to run a minimal example cluster. If you have any idea pls let me know. I will check logs in the instances in the meantime. Also pls give me some feedback on the documentation

### DIFF
--- a/modules/consul-cluster/README.md
+++ b/modules/consul-cluster/README.md
@@ -3,7 +3,7 @@
 This folder contains a [Terraform](https://www.terraform.io/) module to deploy a 
 [Consul](https://www.consul.io/) cluster in [AWS](https://aws.amazon.com/) on top of an Auto Scaling Group. This module 
 is designed to deploy an [Amazon Machine Image (AMI)](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/AMIs.html) 
-that has Consul installed via the [install-consul](../../modules/install-consul) module in this Module.
+that has Consul installed via the [install-consul](https://github.com/hashicorp/terraform-aws-consul/tree/master/modules/install-consul) module in this Module.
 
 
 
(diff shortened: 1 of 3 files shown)


modules/consul-cluster/README.md
```diff
@@ -3,7 +3,7 @@
 This folder contains a [Terraform](https://www.terraform.io/) module to deploy a 
 [Consul](https://www.consul.io/) cluster in [AWS](https://aws.amazon.com/) on top of an Auto Scaling Group. This module 
 is designed to deploy an [Amazon Machine Image (AMI)](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/AMIs.html) 
-that has Consul installed via the [install-consul](https://github.com/hashicorp/terraform-aws-consul/tree/master/modules/install-consul) module in this Module.
+that has Consul installed via the [install-consul](../../modules/install-consul) module in this Module.
```
brikis98: It's irritating, but unfortunately, the Terraform Registry currently requires absolute URLs in READMEs.

modules/consul-cluster/README.md
```diff
@@ -13,25 +13,34 @@ This folder defines a [Terraform module](https://www.terraform.io/docs/modules/u
...
   source = "github.com/hashicorp/terraform-aws-consul//modules/consul-cluster?ref=v0.0.5"
 
   # Specify the ID of the Consul AMI. You should build this using the scripts in the install-consul module.
-  ami_id = "ami-abcd1234"
-  
+  ami_id = "ami-6820c012" # Ubuntu AMi, do not use in producuction
+  instance_type = "t2.medium"
+  cluster_name = "${var.cluster_name}"
+  cluster_size = "${var.num_servers}"
+  ssh_key_name = "${aws_key_pair.auth.key_name}"
```
brikis98: We actually intentionally try to keep the number of parameters small (see the comment below about checking `vars.tf`) or this example code quickly goes out of date as parameters are added/removed/etc. For the full example, it's always best to look at the working example code in the actual `.tf` files.

modules/consul-cluster/README.md
```diff
@@ -13,25 +13,34 @@ This folder defines a [Terraform module](https://www.terraform.io/docs/modules/u
...
+
+  # Configure and start Consul during boot. It will automatically form a cluster with all nodes that have that same tag.
   user_data = <<-EOF
               #!/bin/bash
-              /opt/consul/bin/run-consul --server --cluster-tag-key consul-cluster
+              /opt/consul/bin/run-consul --server --cluster-tag-key demo-consul-cluster
               EOF
-  
+
   # ... See vars.tf for the other parameters you must define for the consul-cluster module
```
brikis98: Ah, this should now say `variables.tf` instead of `vars.tf`.

brikis98:
> I think this is an improvement but I am getting error 500 on the cluster instances

Are you running the example code in the root of this repo? Or something custom?

modules/consul-cluster/README.md
bizmate: Ok I see why they were absolute now. I will change it back

brikis98:
The root example is in the root folder: https://github.com/hashicorp/terraform-aws-consul/blob/master/main.tf

bizmate:
Thanks to you pointing me to the documentation I have managed to run it in my custom adaptation. I think the position of any loose file that is a document would probably be best put under the docs folder instead of root and I also noticed the link to 'consul-cluster example' in  https://github.com/hashicorp/terraform-aws-consul/tree/master/examples/consul-examples-helper is broken. I can change the above if agreed. I also about automating everything. I can write a little command that would spin the cluster and add it do the doc so any user readying the docs can take 5 mins to spin it up

brikis98:
> I think the position of any loose file that is a document would probably be best put under the docs folder instead of root

Unfortunately, the [Terraform Registry](https://registry.terraform.io/) requires that there is code in the root folder. 

>  I also noticed the link to 'consul-cluster example' in https://github.com/hashicorp/terraform-aws-consul/tree/master/examples/consul-examples-helper is broken.

PR to fix that is very welcome!

> I can write a little command that would spin the cluster and add it do the doc so any user readying the docs can take 5 mins to spin it up

What would the command do?

Etiene:
closed due to inactivity

Issue #30, opened 2017-12-03
Consul AMI example - missing Git for AWS example


davidham:
I was trying to set up the AMI for production usage per your [example code on this page](https://github.com/hashicorp/terraform-aws-consul/tree/master/examples/consul-ami). The Amazon Linux version failed with this error:

`amazon-linux-ami: /tmp/script_7892.sh: line 2: git: command not found`

I was able to re-run the Amazon Linux one by modifying the `inline` block as follows:

```
"inline": [
      "sudo yum install -y git",
      "git clone --branch v0.1.0 https://github.com/hashicorp/terraform-aws-consul.git /tmp/terraform-aws-consul",
      "/tmp/terraform-aws-consul/modules/install-consul/install-consul --version {{user `consul_version`}}",
      "/tmp/terraform-aws-consul/modules/install-dnsmasq/install-dnsmasq"
    ],
```

brikis98:
Ah, good point. Ubuntu has git installed by default, but Amazon Linux does not. A PR to improve those docs would be welcome!

jasonmcintosh:
Would you consider this fixed with the dependency list documented?

brikis98:
The example AMI readme now documents this. Thx @jasonmcintosh.

Issue #37, opened 2018-01-19
Init script pointing to /usr/local/bin/supervisor* incorrectly


jasonmcintosh:
the [install-consul](https://github.com/hashicorp/terraform-aws-consul/blob/master/modules/install-consul/install-consul) shell script has:
```
  # On Amazon Linux, /usr/local/bin is not in PATH for the root user, so we add symlinks to /usr/bin, which is in PATH
  if [[ ! -f "/usr/bin/supervisorctl" ]]; then
    sudo ln -s /usr/local/bin/supervisorctl /usr/bin/supervisorctl
  fi
  if [[ ! -f "/usr/bin/supervisord" ]]; then
    sudo ln -s /usr/local/bin/supervisord /usr/bin/supervisord
  fi
```

However, the [init script](https://github.com/hashicorp/terraform-aws-consul/blob/master/modules/install-consul/supervisor-initd-script.sh) is hard coded to a path in /usr/local/bin 
```
supervisorctl=/usr/local/bin/supervisorctl
supervisord=${SUPERVISORD-/usr/local/bin/supervisord}
```
Init script should probably be hard coded to /usr/bin/supervisorctl since the installer script is creating a symlink in /usr/bin - hit this issue trying to install this on a CentOS 7 based AMI.

brikis98:
The init script is one that's used pretty widely (you can see we didn't write it ourselves), so I'm a bit hesitant to change it. We add the symlink mostly for user convenience rather than the init script. On CentOS, does supervisord end up in `/usr/bin` and NOT `/usr/local/bin`?

jasonmcintosh:
Something logic wise....
IF NOT EXIST /usr/local/bin/supervisorctl
IF EXIST /usr/bin/supervisorctl
 -- Link /usr/bin/supervisorctl to /usr/local/bin/supervisorctl
ELSE IF EXIST /usr/local/bin/supervisorctl
 -- Link /usr/local/bin/supervisorctl to /usr/bin/supervisorctl
ELSE
-- ERROR OUT - no supervisorctl found
ENDIF
ENDIF

NOT so sure we even need the /usr/bin/supervisorctl link - since the init script is hard coded to go to /usr/local/bin/supervisorctl.  BUT doesn't hurt I'd guess and that way if something needs to start things, it can do so.

brikis98:
[...]

PR #40, opened 2018-01-26
Fixed symlink issues on CentOS

jasonmcintosh:
Should fix the symlink issues for people.  Does it both directions and then fails the script if it can't find supervisorctl/supervisord

### DIFF
--- a/examples/consul-ami/README.md
+++ b/examples/consul-ami/README.md
@@ -17,6 +17,10 @@ in Consul). To see how to deploy this AMI, check out the [consul-cluster example
 For more info on Consul installation and configuration, check out the 
 [install-consul](https://github.com/hashicorp/terraform-aws-consul/tree/master/modules/install-consul) and [install-dnsmasq](https://github.com/hashicorp/terraform-aws-consul/tree/master/modules/install-dnsmasq) documentation.
 
+## Dependencies
+1.  AWSCLI must be installed on the base AMI in order for run-consul to run
+1.  PIP must be installed on the base AMI in order to install supervisord
+1.  Git CAN be installed on the base AMI if you want to use clone commands
 
 
 ## Quick start
--- a/modules/install-consul/install-consul
+++ b/modules/install-consul/install-consul
@@ -78,8 +78,8 @@ function install_supervisord_debian {
 }
 
 function two_way_symlink() {
-  src=$1
-  dest=$2
+  local src="$1"
+  local dest="$2"
 
   if [[ ! -f "$dest" ]] && [[ ! -f "$src" ]]; then
     echo "Missing source AND destination... exiting..."
--- a/modules/install-consul/install-consul
+++ b/modules/install-consul/install-consul
@@ -99,14 +99,14 @@ function install_supervisord_amazon_linux {
 
   if [[ ! -f "/usr/local/bin/supervisord" ]]; then
     if [[ -f "/usr/bin/supervisord" ]]; then
-      sudo ln -s /usr/bin/supervisorctl /usr/local/bin/supervisorctl
+      sudo ln -s /usr/bin/supervisord /usr/local/bin/supervisord
     else
       echo "Unable to find supervisord... exiting"
       exit -5
     fi
   elif [[ ! -f "/usr/bin/supervisord" ]]; then
     if [[ -f "/usr/local/bin/supervisord" ]]; then
-      sudo ln -s /usr/local/bin/supervisorctl /usr/bin/supervisorctl
+      sudo ln -s /usr/local/bin/supervisord /usr/bin/supervisord
     fi
   fi
 
--- a/modules/install-consul/install-consul
+++ b/modules/install-consul/install-consul
@@ -82,11 +82,32 @@ function install_supervisord_amazon_linux {
   sudo pip install supervisor
 
   # On Amazon Linux, /usr/local/bin is not in PATH for the root user, so we add symlinks to /usr/bin, which is in PATH
-  if [[ ! -f "/usr/bin/supervisorctl" ]]; then
-    sudo ln -s /usr/local/bin/supervisorctl /usr/bin/supervisorctl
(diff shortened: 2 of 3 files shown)


examples/consul-ami/README.md
```diff
@@ -17,6 +17,10 @@ in Consul). To see how to deploy this AMI, check out the [consul-cluster example
 For more info on Consul installation and configuration, check out the 
 [install-consul](https://github.com/hashicorp/terraform-aws-consul/tree/master/modules/install-consul) and [install-dnsmasq](https://github.com/hashicorp/terraform-aws-consul/tree/master/modules/install-dnsmasq) documentation.
 
+## Dependencies
+1.  AWSCLI must be installed on the base AMI in order for run-consul to run
+1.  PIP must be installed on the base AMI in order to install supervisord
+1.  Git CAN be installed on the base AMI if you want to use clone commands
```
brikis98: Thanks for adding 👍

modules/install-consul/install-consul
```diff
@@ -82,11 +82,32 @@ function install_supervisord_amazon_linux {
...
+      sudo ln -s /usr/local/bin/supervisorctl /usr/bin/supervisorctl
+
+    fi
   fi
-  if [[ ! -f "/usr/bin/supervisord" ]]; then
-    sudo ln -s /usr/local/bin/supervisord /usr/bin/supervisord
+
+  if [[ ! -f "/usr/local/bin/supervisord" ]]; then
+    if [[ -f "/usr/bin/supervisord" ]]; then
+      sudo ln -s /usr/bin/supervisorctl /usr/local/bin/supervisorctl
```
brikis98: Shouldn't this be symlinking `supervisord`?

modules/install-consul/install-consul
```diff
@@ -82,11 +82,32 @@ function install_supervisord_amazon_linux {
...
+    else
+      echo "Unable to find supervisord... exiting"
+      exit -5
+    fi
+  elif [[ ! -f "/usr/bin/supervisord" ]]; then
+    if [[ -f "/usr/local/bin/supervisord" ]]; then
+      sudo ln -s /usr/local/bin/supervisorctl /usr/bin/supervisorctl
```
brikis98: Same comment here

jasonmcintosh: Bleh good catch!  Really should have abstracted it to a function.  Will fix :) I'll blame baby brain!

modules/install-consul/install-consul
```diff
@@ -82,11 +82,32 @@ function install_supervisord_amazon_linux {
...
+  if [[ ! -f "/usr/local/bin/supervisorctl" ]]; then
+    if [[ -f "/usr/bin/supervisorctl" ]]; then
+      sudo ln -s /usr/bin/supervisorctl /usr/local/bin/supervisorctl
+    else
+      echo "Unable to find supervisorctl tool... exiting"
+      exit -5
+    fi
+  elif [[ ! -f "/usr/bin/supervisorctl" ]]; then
+    # IF supervisorctl does NOT exist in /usr/bin, check for it in /usr/local/bin and link it to /usr/bin
+    if [[ -f "/usr/local/bin/supervisorctl" ]]; then
```
brikis98: Should there be an `else` clause here with an error? Same with the if-statements below.

This logic can be tricky to follow, so perhaps extract it into a `two_way_symlink(src, dst)` function that (a) creates a symlink from `src` -> `dst` if `dst` exists and `src` doesn't (b) creates a symlink from `dst` -> `src` if `src` exists and `dst` doesn't, and (c) exits with an error if neither exist.

jasonmcintosh:
This should also fix https://github.com/hashicorp/terraform-aws-consul/issues/30 btw... I added a "Dependencies" block.

modules/install-consul/install-consul
jasonmcintosh: Should be better :)

modules/install-consul/install-consul
```diff
@@ -77,17 +77,33 @@ function install_supervisord_debian {
   sudo systemctl enable supervisor
 }
 
+function two_way_symlink() {
+  src=$1
+  dest=$2
```
brikis98: We follow the [google shell style guide](https://google.github.io/styleguide/shell.xml), so all function params should be `local readonly` and wrap the argument in double quotes (e.g., `"$1"`).

modules/install-consul/install-consul
```diff
@@ -77,17 +77,33 @@ function install_supervisord_debian {
...
+
+  if [[ ! -f "$dest" ]] && [[ ! -f "$src" ]]; then
+    echo "Missing source AND destination... exiting..."
```
brikis98: Please print out `$src` and `$dest` here to help with understanding what the error is referring to.

modules/install-consul/install-consul
```diff
@@ -77,17 +77,33 @@ function install_supervisord_debian {
...
 
   # On Amazon Linux, /usr/local/bin is not in PATH for the root user, so we add symlinks to /usr/bin, which is in PATH
-  if [[ ! -f "/usr/bin/supervisorctl" ]]; then
-    sudo ln -s /usr/local/bin/supervisorctl /usr/bin/supervisorctl
-  fi
-  if [[ ! -f "/usr/bin/supervisord" ]]; then
-    sudo ln -s /usr/local/bin/supervisord /usr/bin/supervisord
-  fi
+  two_way_symlink "/usr/bin/supervisorctl" "/usr/local/bin/supervisorctl"
+  two_way_symlink "/usr/bin/supervisord" "/usr/local/bin/supervisord"
```
brikis98: Ah, this is much easier to understand now, thx!

jasonmcintosh:
https://github.com/hashicorp/terraform-aws-consul/issues/37  

The init script is hard coded to /usr/local/bin locations.  The install-consul script had no location so used path, which then failed.  This allows both the init script to function and installer to function.

PR #45, opened 2018-02-12
Added Support for EC2 Spot Instances

clovis818:


brikis98:
Thanks for the PR! Code changes look good. How did you test this?

clovis818:
I haven't tested this out yet, but based on Terraform aws_launch_configuration its supported.
https://www.terraform.io/docs/providers/aws/r/launch_configuration.html#spot_price

brikis98:
Yea, it looks reasonable, but we can't accept any PRs without tests, and we can't run tests for external PRs automatically for security reasons. If you could fire up a cluster with/without spot pricing and use the helper script to check everything still runs, I'd be grateful!

clovis818:
Fixed a few bugs and Tested both consul with and without spot pricing.

1. Added a 30 second sleep to user-data-client.sh and user-data-server.sh
The aws_autoscaling_group did not propagate tags until after the servers launched, which caused run-consul to generate multiple configs with "bootstrap_expect":"1". 

2.  The aws_autoscaling_group tenancy has to be be set to null for spot instances.
AWS will throw an error if its set to "default". 
The VPC Tenancy will be used when its null.
https://docs.aws.amazon.com/autoscaling/ec2/userguide/asg-in-vpc.html#as-vpc-tenancy

brikis98:
> Added a 30 second sleep to user-data-client.sh and user-data-server.sh
> The aws_autoscaling_group did not propagate tags until after the servers launched, which caused run-consul to generate multiple configs with "bootstrap_expect":"1".

Hmm, I have not seen this issue before. Is this something you hit specifically with spot instances?

Otherwise, I'm not sure why you would get `"bootstrap_expect" : 1`, as we are looking up the ASG's [DesiredCapacity](https://github.com/hashicorp/terraform-aws-consul/blob/master/modules/run-consul/run-consul#L117-L118), not the actual number of servers running in it. 

> The launch_configuration tenancy has to be be set to null for spot instances. AWS will throw an error if its set to "default". The VPC Tenancy will be used when its null.

Ah, right, good catch. This is why we ask to test :)

examples/consul-examples-helper/consul-examples-helper.sh
```diff
@@ -152,16 +152,20 @@ function get_consul_cluster_ips {
 
 function print_instructions {
   local readonly server_ips=($@)
-  local readonly server_ip="${server_ips[0]}"
+  local readonly server_ip_0="${server_ips[0]}"
+  local readonly server_ip_1="${server_ips[1]}"
+  local readonly server_ip_2="${server_ips[2]}"
```
brikis98: Why this change?

examples/consul-examples-helper/consul-examples-helper.sh
```diff
@@ -152,16 +152,20 @@ function get_consul_cluster_ips {
...
   instructions+=("\nYour Consul servers are running at the following IP addresses:\n\n${server_ips[@]/#/    }\n")
   instructions+=("Some commands for you to try:\n")
-  instructions+=("    consul members -http-addr=$server_ip:8500")
-  instructions+=("    consul kv put -http-addr=$server_ip:8500 foo bar")
-  instructions+=("    consul kv get -http-addr=$server_ip:8500 foo")
+  instructions+=("    consul members -http-addr=$server_ip_0:8500")
+  instructions+=("    consul kv put -http-addr=$server_ip_0:8500 foo bar")
+  instructions+=("    consul kv get -http-addr=$server_ip_0:8500 foo")
+  instructions+=("    consul kv get -http-addr=$server_ip_1:8500 foo")
+  instructions+=("    consul kv get -http-addr=$server_ip_2:8500 foo")
```
brikis98: Seems more sane to do this in a for-loop, rather than a hard-coded list of 3 servers.

modules/consul-cluster/variables.tf
```diff
@@ -93,9 +93,14 @@ variable "associate_public_ip_address" {
...
 
+variable "spot_price" {
+  description = "The maximum hourly price to pay for EC2 Spot Instances."
+  default     = ""
+}
+
 variable "tenancy" {
-  description = "The tenancy of the instance. Must be one of: default or dedicated."
-  default     = "default"
+  description = "The tenancy of the instance. Must be one of: null, default or dedicated. For EC2 Spot Instances only null or dedicated can be used."
```
brikis98: To be a bit clearer, instead of "null" I'd say "empty string."

clovis818:
When the servers launch tags are not applied for a second or two, which can cause a race condition with run-consul. Its has happened to me several times.

examples/consul-examples-helper/consul-examples-helper.sh
clovis818: To verify that consul has elected a leader and servers have synchronized.

brikis98:
Ah, I see. Are you seeing this warning in your logs? https://github.com/hashicorp/terraform-aws-consul/blob/master/modules/run-consul/run-consul#L128

Also, if this is indeed the race condition, where the instance looks up its own tags before they have propagated, then the fix is not a random sleep, but to fetch the tags in a retry loop until the ASG tag is found (or some max timeout is reached).

clovis818:
How about instead of the 30 second sleep using:
`aws ec2 wait instance-status-ok`
https://docs.aws.amazon.com/cli/latest/reference/ec2/wait/instance-status-ok.html
`aws ec2 wait system-status-ok`
https://docs.aws.amazon.com/cli/latest/reference/ec2/wait/system-status-ok.html

brikis98:
I'm not sure either of those implies tags have propagated, so a for-loop to fetch the tags until they are present, with a sleep in between, is probably safer.

clovis818:
Okay added the for-loop to fetch tags.

Server Logs:
https://pastebin.com/LY5irrU3

Client Logs:
https://pastebin.com/RfMw8Gcz

modules/run-consul/run-consul
```diff
@@ -103,9 +106,25 @@ function get_instance_region {
...
+    if [[ "$count_tags" -gt 0 ]]; then
+      log_info "This Instance $instance_id in $instance_region has Tags."
+      echo "$tags"
+      return
+    else
+      log_warn "This Instance $instance_id in $instance_region does not have any Tags."
+      log_warn "Will sleep for $SLEEP_BETWEEN_RETRIES_SEC seconds and try again."
+      sleep "$SLEEP_BETWEEN_RETRIES_SEC"
+    fi
+  done
```
brikis98: After this `for` loop completes, you should `log_error` and exit with an error code to indicate tags could not be found after `MAX_RETRIES` retries.

brikis98:
https://github.com/hashicorp/terraform-aws-consul/releases/tag/v0.2.1

PR #47, opened 2018-02-21
Add tags variable to allow passing in extra tags

miguelaferreira:
This PR adds a new variable `tags` to pass in extra tag blocks for the autoscaling group, and via tag propagation (if enabled) also to the instances.

My use case is building a complete environment of which a consul cluster is part. All resources in that environment need to be tagged with custom tags. Examples of such tags are the name of the environment and a flag to determine if monitoring should be enabled on those resources.

I've tested this PR using test-kitchen and awspec.

I've specified tags like this:
```hcl
module "consul_cluster" {
  source = "github.com/miguelaferreira/terraform-aws-consul//modules/consul-cluster?ref=extra-tags"

  ...

  tags = [
    {
      key                 = "Project"
      value               = "${var.project}"
      propagate_at_launch = true
    },
    {
      key                 = "Environment"
      value               = "${var.environment}"
      propagate_at_launch = true
    },
    {
      key                 = "Monitoring"
      value               = "${var.monitoring_enabled ? "true" : "false"}"
      propagate_at_launch = true
    },
  ]
}
```

And I have an awspec test that looks like this:
```ruby
require 'awspec'
require 'rubygems'

tf_output = eval(ENV['KITCHEN_KITCHEN_TERRAFORM_OUTPUT'])

consul_autoscaling_group_name = tf_output[:consul_autoscaling_group_name][:value]

describe autoscaling_group(consul_autoscaling_group_name) do
  it { should exist }
  it { should have_tag('Project').value('project') }
  it { should have_tag('Environment').value('test') }
  it { should have_tag('Monitoring').value('true') }
end
```

The test results were:
```
autoscaling_group 'NloGatewayTestConsul20180221102548649600000011'
  should exist
  should have tag "Project"
  should have tag "Environment"
  should have tag "Monitoring"

Finished in 0.29098 seconds (files took 1.89 seconds to load)
4 examples, 0 failures
```

### DIFF
--- a/modules/consul-cluster/main.tf
+++ b/modules/consul-cluster/main.tf
@@ -28,17 +28,19 @@ resource "aws_autoscaling_group" "autoscaling_group" {
   health_check_grace_period = "${var.health_check_grace_period}"
   wait_for_capacity_timeout = "${var.wait_for_capacity_timeout}"
 
-  tag {
-    key                 = "Name"
-    value               = "${var.cluster_name}"
-    propagate_at_launch = true
-  }
-
-  tag {
-    key                 = "${var.cluster_tag_key}"
-    value               = "${var.cluster_tag_value}"
-    propagate_at_launch = true
-  }
+  tags = [
+    {
+      key                 = "Name"
+      value               = "${var.cluster_name}"
+      propagate_at_launch = true
+    },
+    {
+      key                 = "${var.cluster_tag_key}"
+      value               = "${var.cluster_tag_value}"
+      propagate_at_launch = true
+    },
+    "${var.tags}",
+  ]
 }
 
 # ---------------------------------------------------------------------------------------------------------------------
--- a/modules/consul-cluster/variables.tf
+++ b/modules/consul-cluster/variables.tf
@@ -184,3 +184,9 @@ variable "ssh_port" {
   description = "The port used for SSH connections"
   default     = 22
 }
+
+variable "tags" {
+  description = "List fo extra tag blocks added to the autoscaling group configuration"
+  type        = "list"
+  default     = []
+}
--- a/modules/consul-cluster/variables.tf
+++ b/modules/consul-cluster/variables.tf
@@ -186,7 +186,7 @@ variable "ssh_port" {
 }
 
 variable "tags" {
-  description = "List fo extra tag blocks added to the autoscaling group configuration"
+  description = "List fo extra tag blocks added to the autoscaling group configuration. Each element in the list is a map containing keys 'key', 'value', and 'propagate_at_launch' mapped to the respective values."
   type        = "list"
   default     = []
 }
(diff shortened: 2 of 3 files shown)


modules/consul-cluster/main.tf
```diff
@@ -28,17 +28,19 @@ resource "aws_autoscaling_group" "autoscaling_group" {
...
+      key                 = "Name"
+      value               = "${var.cluster_name}"
+      propagate_at_launch = true
+    },
+    {
+      key                 = "${var.cluster_tag_key}"
+      value               = "${var.cluster_tag_value}"
+      propagate_at_launch = true
+    },
+    "${var.tags}",
```
brikis98: Does this actually work? I'd expect you to have to call `concat`.

modules/consul-cluster/variables.tf
```diff
@@ -184,3 +184,9 @@ variable "ssh_port" {
   description = "The port used for SSH connections"
   default     = 22
 }
+
+variable "tags" {
+  description = "List fo extra tag blocks added to the autoscaling group configuration"
```
brikis98: Please mention that each block needs to specify `key`, `value`, and `propagate_at_launch`.

modules/consul-cluster/variables.tf
```diff
@@ -184,3 +184,9 @@ variable "ssh_port" {
...
+  type        = "list"
+  default     = []
```
brikis98: Could you add an example? E.g.

```hcl
  default = []
  # Example:
  #
  # default = [{ ... } ]
```

modules/consul-cluster/main.tf
miguelaferreira: Yes it does, I've tested this by pulling my fork of the module to an existing infrastructure configuration and the tags got defined correctly. I considered using `concat` but found this construction to read better.

modules/consul-cluster/variables.tf
miguelaferreira: Will do

modules/consul-cluster/variables.tf
miguelaferreira: I did add an example in the `main.tf` (in the root of the repository). Where else would you like to have it in?

miguelaferreira:
I've mentioned this in a reply to the review, but I'll mention it again here. I've tested this PR by pulling in code form this branch into a module of my own that uses this (`consul-cluster`) module. Then in my CI builds I've verified the tags were present in the AWS autoscaling group. I'm in the process of applying this branch to an existing infrastructure that was built using `v0.1.1`of this module. I don't expect that this change will break the "terraform cycle" on my infrastructure since its only adding more tags.

modules/consul-cluster/main.tf
brikis98: Ah, neat, thx!

brikis98:
https://github.com/hashicorp/terraform-aws-consul/releases/tag/v0.1.2

PR #48, opened 2018-02-21
Use variable cluster_name as a name prefix for the autoscaling group

miguelaferreira:
This PR reuses the `cluter_name` variable to set `name_prefix` on the cluster autoscaling group resource. 

As the documentation mentions, the `cluster_name` should be "*used to namespace all resources created by this module*".

I've tested this PR with test-ktichen and awspec. I've defined the name of the cluster like this:
```hcl
module "consul_cluster" {
  source = "github.com/miguelaferreira/terraform-aws-consul//modules/consul-cluster?ref=namespace-autoscaling-group"

  cluster_name = "ProjectTestConsul"

  ...
}
```

I wrote a awspec test like this:
```ruby
require 'awspec'
require 'rubygems'

tf_output = eval(ENV['KITCHEN_KITCHEN_TERRAFORM_OUTPUT'])

consul_autoscaling_group_name = tf_output[:consul_autoscaling_group_name][:value]

describe consul_autoscaling_group_name do
  it { should start_with 'ProjectTestConsul' }
end
```

The results were:
```
ProjectTestConsul20180221110521541200000011
  should start with "ProjectTestConsul"

...

Finished in 0.17729 seconds (files took 1.85 seconds to load)
5 examples, 0 failures
```

### DIFF
--- a/modules/consul-cluster/main.tf
+++ b/modules/consul-cluster/main.tf
@@ -11,6 +11,8 @@ terraform {
 # ---------------------------------------------------------------------------------------------------------------------
 
 resource "aws_autoscaling_group" "autoscaling_group" {
+  name_prefix = "${var.cluster_name}"
+
   launch_configuration = "${aws_launch_configuration.launch_configuration.name}"
 
   availability_zones  = ["${var.availability_zones}"]


brikis98:
Code change LGTM. How did you test it?

miguelaferreira:
@brikis98 I've updated the PR with the testing info. Thanks for reviewing so quickly 👍

miguelaferreira:
@brikis98: any chance you could give this PR another look?

brikis98:
Oh weird, I somehow didn't get a notification for your previous comment. Thx for the reminder. Code and test approach LGTM, thx!

brikis98:
https://github.com/hashicorp/terraform-aws-consul/releases/tag/v0.2.0

kostyrev:
Release notes state that
>Terraform treats renaming an ASG as deleting the old one and creating a new one, so be careful when updating!

Any advice how to be careful?  
`terraform apply` *will* destroy all of your nodes with the data.
Is `careful` means backup your consul and update with downtime?

brikis98:
Probably the best option is to bring up a new ASG with the `consul-cluser` module, let the nodes sync up to the existing one and replicate the data, and then delete the old one. I'd recommend testing in pre-prod first to make sure the cluster handles the addition and removal of nodes properly!

miguelaferreira:
@kostyrev wrapping the ASG in a cloudformation stack allows you to set an update policy, and one of the types is `AutoScalingRollingUpdate`, which will update each instance at a time. It also gives possibilities to roll back updates. I use this approach to [manage ECS clusters with terraform](https://github.com/miguelaferreira/terraform-aws-ecs-cluster/blob/master/main.tf#L207).

kostyrev:
@miguelaferreira but terraform considers this a new resource so I will delete it anyway

miguelaferreira:
I'm not absolutely sure about that, because the resource terraform controls is the cloudformation stack, not the ASG. But I would test it first ;)

eedwards-sk:
@miguelaferreira Can we fix this to append a dash, as well, similar to the launch configuration?

example launch configuration name:

```
cluster_name-20181024222144046700000004
```

current example autoscaling group name:

```
cluster_name20181024222144046700000004
```

proposed example autoscaling group name:

```
cluster_name-20181024222144046700000004
```

miguelaferreira:
Sure @eedwards-sk, let me open a new PR

PR #50, opened 2018-03-02
Specify custom datacenter in run-consul script

icy-arctic-fox:
This PR adds the ability to specify a datacenter field in the Consul config. As it exists now, the datacenter is set to the AWS region name. In our situation, we can have multiple deployments in the same region (staging vs. production).

A `--datacenter` option has been added to the run-consul script. It is optional and defaults to the old behavior of using the AWS region.

**Downtime needed:** No.
This change does not require downtime. New nodes can be swapped in with this change and there are no adverse effects. Specifying a different datacenter may require downtime, but it is the same as-if changing the datacenter manually.

**Backwards compatible:** Yes.
The default behavior without the option is to use the region name, which is identical to what it was before the change.

The change was tested in our AWS account with and without the --datacenter option and it worked as expected.

### DIFF
--- a/modules/run-consul/README.md
+++ b/modules/run-consul/README.md
@@ -59,6 +59,7 @@ The `run-consul` script accepts the following arguments:
   in `--cluster-tag-value`.
 * `cluster-tag-value` (optional): Automatically form a cluster with Instances that have the tag key in 
   `--cluster-tag-key` and this tag value.
+* `datacenter` (optional): The name of the datacenter the cluster reports. Default is the AWS region name.
 * `config-dir` (optional): The path to the Consul config folder. Default is to take the absolute path of `../config`, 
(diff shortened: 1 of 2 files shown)


brikis98:
Thanks for the PR! Changes LGTM. 

One thing worth mentioning for the future: it's easy to override the default config, even for settings the `run-consul` script doesn't expose directly, [as documented here](https://github.com/hashicorp/terraform-aws-consul/tree/master/modules/run-consul#overriding-the-configuration).

brikis98:
https://github.com/hashicorp/terraform-aws-consul/releases/tag/v0.2.2

icy-arctic-fox:
Thanks for merging!

I was conflicted about using a config file. The only additional customization we needed was the datacenter. Everything else could be done with what's already there.

PR #57, opened 2018-03-19
Using `aws_autoscaling_attachment` instead

### LINKED ISSUE #55: Using `aws_autoscaling_attachment`
I would like to suggest the same changes as in https://github.com/hashicorp/terraform-aws-nomad/issues/19.

For this case, I would like to attach ELBs to access Consul's HTTP endpoint and the UI after the creation of the ASG in a separate module.

May I know your thoughts on this?

lawliet89:
Fixes #55


brikis98:
https://github.com/hashicorp/terraform-aws-consul/releases/tag/v0.3.0

Issue #41, opened 2018-01-29
RFE: Support TLS/Encrypt options


jasonmcintosh:
Opening this as a reference - I plan on working to add this at some point, as we're wanting this when using it for Vault purposes.  

Documentation:
https://www.consul.io/docs/agent/encryption.html

Note there WOULD be a challenge if someone attempted to use this on an existing cluster that did NOT have encryption appropriately enabled, but that's a separate discussion.

Places that need to be modified:
Config file:
https://github.com/hashicorp/terraform-aws-consul/blob/master/modules/run-consul/run-consul#L194
User data files to pass it to run-consul
https://github.com/hashicorp/terraform-aws-consul/tree/master/examples/root-example
Finally the below to add variables for the user data scripts
https://github.com/hashicorp/terraform-aws-consul/blob/master/main.tf#L92

Might be missing things... but it's a start

brikis98:
Does this cover what you're looking for? https://github.com/hashicorp/terraform-aws-consul/tree/master/modules/run-consul#how-do-you-handle-encryption

jasonmcintosh:
Exactly :)  The vault module has support for these on AWS - be nice to add it here.  Figure I'll work on that a bit here when I get some spare time - doesn't look like it'll be too hard to add support for it.

patoarvizu:
I'll be submitting a PR for this shortly...

patoarvizu:
I'd appreciate any feedback on the PR above :)

brikis98:
Fixed in #58.

PR #58, opened 2018-04-06
Enable encryption

patoarvizu:
This should address #41.

Modified `run-consul` to add 5 optional flags: `----enable-gossip-encryption`, `--gossip-encryption-key`, `--enable-rpc-encryption`, `--ca-file-path`, `--cert-file-path` and `--key-file-path`. As their names imply, they're used to enable gossip and/or RPC encryption, as per the [Consul encryption documentation](https://www.consul.io/docs/agent/encryption.html).

I also added a separate example with sample certificates (and a note that they're insecure and shouldn't be used in production). One thing I debated about adding or not was hard-coding a sample gossip encryption key in the example. I decided against it, but I can add it if you think that would be useful or would make it easier to run the example

For what it's worth, I was able to upgrade an existing Vault/Cluster setup with no encryption on the Consul agents to using both gossip and RPC encryption with no downtime using the changes I'm proposing and the process outlined in the Consul encryption document linked above.

Let me know if there's anything you feel I should change or add.


patoarvizu:
Mmhh... now that I'm thinking about it, I might want to switch to use `ca_path` instead of `ca_file`. I'll add that change, but I'd still welcome feedback for the change in general.

examples/example-with-encryption/README.md
```diff
@@ -0,0 +1,23 @@
+# Consul cluster with encryption example
+
+This folder contains an example of a Packer manifest that creates an AMI with the required certs for TLS validation, as well as installing an updated version of the `run-consul` script that accepts parameters for enabling RPC and/or gossip encryption.
```
brikis98: You should mention that this folder also contains Terraform configurations that can be used to deploy the AMI to AWS.

examples/example-with-encryption/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+The resulting AMI id can then be passed as a parameter to `variables.tf`. The `enable_gossip_encryption` and `enable_rpc_encryption` variables are set to `true` by default in this example, but they don't have to be in your implementation. In this example they're passed as parameters to the `user_data` template to generate the flags passed to `run-consul` but you can use a different strategy.
+
+The end result of this example should be a cluster of 3 Consul servers and 3 Consul clients, all running on individual EC2 instances. If the default variables are used, both gossip and RPC encryption will be enabled. You can validate this by trying to bring up another Consul node or cluster NOT running with encryption and attempt to join the existing cluster.
+
+Running this example with encryption turned off and then attempt to upgrade it to use encryption is a good exercise to validate that a production cluster can be upgraded with minimal impact.
+
+To understand more about how Consul handles encryption or how you can upgrade to use encryption without downtime, check out the [Consul encryption documentation](https://www.consul.io/docs/agent/encryption.html).
```
brikis98: Please add a big BOLD warning that says the certs in this repo are JUST for learning/testing and should NOT be used in prod! I know the TLS certs folder's README says this, but we should have this warning in as many places as possible :)

examples/example-with-encryption/main.tf
```diff
@@ -0,0 +1,140 @@
+# ---------------------------------------------------------------------------------------------------------------------
+# DEPLOY A CONSUL CLUSTER IN AWS
+# These templates show an example of how to use the consul-cluster module to deploy Consul in AWS. We deploy two Auto
+# Scaling Groups (ASGs): one with a small number of Consul server nodes and one with a larger number of Consul client
+# nodes. Note that these templates assume that the AMI you provide via the ami_id input variable is built from
+# the examples/consul-ami/consul.json Packer template.
```
brikis98: Comment doesn't match up with code (e.g., wrong Packer template).

examples/example-with-encryption/main.tf
```diff
@@ -0,0 +1,140 @@
...
+# ---------------------------------------------------------------------------------------------------------------------
+
+data "template_file" "user_data_server" {
+  template = "${file("${path.module}/user-data-server.sh")}"
+
+  vars {
+    cluster_tag_key   = "${var.cluster_tag_key}"
+    cluster_tag_value = "${var.cluster_name}"
+    gossip_encryption_configuration = "${var.enable_gossip_encryption && var.gossip_encryption_key != "" ? "--enable-gossip-encryption --gossip-encryption-key ${var.gossip_encryption_key}" : ""}"
+    rpc_encryption_configuration = "${var.enable_rpc_encryption && var.ca_path != "" && var.cert_file_path != "" && var.key_file_path != "" ? "--enable-rpc-encryption --ca-path ${var.ca_path} --cert-file-path ${var.cert_file_path} --key-file-path ${var.key_file_path}" : ""}"
```
brikis98: I'm not a big fan of passing command-line args from Terraform to user data. I think it's a bit cleaner if you *just* pass the parameters straight through:

```hcl
enable_gossip_encryption = "${var.enable_gossip_encryption}"
gossip_encryption_key    = "${var.gossip_encryption_key}"

enable_rpc_encryption = "${var.enable_rpc_encryption}"
ca_path               = "${var.ca_path}"

# ... etc ...
```

The User Data script itself can check the values of the parameters and set command line args as necessary.

examples/example-with-encryption/packer/README.md
```diff
@@ -0,0 +1,16 @@
+# Example TLS Certificates
```
brikis98: You may want to mention how you created these so users know how to create their own!

examples/example-with-encryption/packer/consul-with-certs.json
```diff
@@ -0,0 +1,84 @@
...
+    "type": "shell",
+    "inline": [
+      "sudo mkdir -p /opt/consul/tls/ca",
+      "sudo mv /tmp/ca.crt.pem /opt/consul/tls/ca/",
+      "sudo mv /tmp/consul.crt.pem /opt/consul/tls/",
+      "sudo mv /tmp/consul.key.pem /opt/consul/tls/",
+      "sudo chown -R consul:consul /opt/consul/tls/*",
+      "sudo chmod 600 /opt/consul/tls/ca/ca.crt.pem",
+      "sudo chmod 600 /opt/consul/tls/consul.crt.pem",
+      "sudo chmod 600 /opt/consul/tls/consul.key.pem"
```
brikis98: Hm, I wonder if `install-consul` should take in `--ca-path`, `cert-path`, etc params? If you set them, it could do this copying and `chmod`ing for you, which seems worthwhile.

examples/example-with-encryption/user-data-client.sh
```diff
@@ -0,0 +1,15 @@
...
+# built from the Packer template in examples/consul-ami/consul.json.
+
+set -e
+
+# Send the log output from this script to user-data.log, syslog, and the console
+# From: https://alestic.com/2010/12/ec2-user-data-output/
+exec > >(tee /var/log/user-data.log|logger -t user-data -s 2>/dev/console) 2>&1
+
+# These variables are passed in via Terraform template interplation
+/opt/consul/bin/run-consul --client --cluster-tag-key "${cluster_tag_key}" --cluster-tag-value "${cluster_tag_value}" ${gossip_encryption_configuration} ${rpc_encryption_configuration}
```
brikis98: Please add a comment that says something along the lines of "to keep this example simple, we are passing  the gossip encryption key in plain text to User Data, but in production usage, it should always be encrypted using a tool like Vault or KMS, and only decrypted, in memory, just before usage."

examples/example-with-encryption/variables.tf
```diff
@@ -0,0 +1,87 @@
...
+# AWS_ACCESS_KEY_ID
+# AWS_SECRET_ACCESS_KEY
+
+# ---------------------------------------------------------------------------------------------------------------------
+# OPTIONAL PARAMETERS
+# These parameters have reasonable defaults.
+# ---------------------------------------------------------------------------------------------------------------------
+
+variable "ami_id" {
+  description = "The ID of the AMI to run in the cluster. This should be an AMI built from the Packer template under examples/consul-ami/consul.json. To keep this example simple, we run the same AMI on both server and client nodes, but in real-world usage, your client nodes would also run your apps. If the default value is used, Terraform will look up the latest AMI build automatically."
```
brikis98: Wrong Packer template path

examples/example-with-encryption/variables.tf
```diff
@@ -0,0 +1,87 @@
...
+  default = true
+}
+
+variable "enable_rpc_encryption" {
+  description = "Encrypt RPC traffic between nodes. Must also specify TLS certificates and keys."
+  default = true
+}
+
+variable "gossip_encryption_key" {
+  description = "16 byte cryptographic key to encrypt gossip traffic between nodes. Must set 'enable_gossip_encryption' to true for this to take effect."
```
brikis98: Please add a comment here along the lines of "WARNING: to keep this example simple, we set the encryption key in plaintext. That means this secret will be stored in plaintext both in version control and in your Terraform state. In production usage, you should instead encrypt the secret using a tool like Vault or KMS, and have your User Data decrypt it just before usage."

examples/example-with-encryption/variables.tf
```diff
@@ -0,0 +1,87 @@
...
+  default = ""
```
brikis98: Either there should be no default to force a user to specify the key or you should put some default in here like "example-secret-just-for-testing".

modules/run-consul/README.md
```diff
@@ -65,6 +65,12 @@ The `run-consul` script accepts the following arguments:
...
   relative to the `run-consul` script itself.
 * `user` (optional): The user to run Consul as. Default is to use the owner of `config-dir`.
+* `enable-gossip-encryption` (optional): Enable encryption of gossip traffic between nodes. If set, you must also specify `gossip-encryption-key`.
+* `gossip-encryption-key` (optional): The key to use for encrypting gossip traffic. Must be specified with `enable-gossip-encryption`.
+* `enable-rpc-encryption` (optional): Enable encryption of RPC traffic between nodes. Must also specify `ca-file-path`, `cert-file-path` and `key-file-path`.
+* `ca-file-path` (optional): Path to the CA file used to verify outgoing connections. Must be specified with `enable-rpc-encryption`, `cert-file-path` and `key-file-path`.
+* `cert-file-path` (optional): Path to the certificate file used to verify incoming connections. Must be specified with `enable-rpc-encryption`, `ca-file-path`, and `key-file-path`.
+* `key-file-path` (optional): Path to the certificate key used to verify incoming connections. Must be specified with `enable-rpc-encryption`, `ca-file-path` and `cert-file-path`.
```
brikis98: I'm guessing the cert/key aren't used to "verify" incoming connections, but to encrypt them, right?

modules/run-consul/README.md
```diff
@@ -193,6 +199,8 @@ configuration file (e.g. `encryption.json`) in the Consul config dir (default lo
 }
 ```
 
+Alternatively, you can run `run-consul` with `--enable-gossip-encryption` and `--gossip-encryption-key` to have this script
+automatically generate the gossip encryption settings in `/opt/consul/config/gossip-encryption.json`.
```
brikis98: I'd actually list this option as the primary way to do it, and make the more manual approach above the "alternative". Same for RPC encryption.

modules/run-consul/run-consul
```diff
@@ -173,6 +181,42 @@ function assert_is_installed {
   fi
 }
 
+function generate_rpc_encryption_config {
```
brikis98: There's no need to write the RPC or gossip encryption configs to a separate config file. In fact, it makes it a bit harder for users to override the defaults. I'd write it to the same config as the rest of the configs written by `run-consul`.

modules/run-consul/run-consul
```diff
@@ -329,6 +373,32 @@ function run {
         datacenter="$2"
         shift
         ;;
+      --enable-gossip-encryption)
+        enable_gossip_encryption="true"
```
brikis98: `enable_gossip_encryption` and all the other vars should be declared above as `local` vars with reasonable defaults (`false` for the bools, empty for the strings).

modules/run-consul/run-consul
patoarvizu: I did it this way because I felt like this would reduce complexity by not creating additional `if`/`else` code paths in the same `generate_consul_config`, but I'm ok with doing it that way if that's your preference.

modules/run-consul/run-consul
patoarvizu: Will do.

modules/run-consul/README.md
patoarvizu: Mmmhh... actually the official documentation uses "verify", and in fact the configuration fields are called `verify_ougoing` and `verify_incoming`, https://www.consul.io/docs/agent/encryption.html. I'm not so sure what verb to use...

brikis98: Well, I guess it's actually both verify and encrypt, depending on whether `verify_outgoing` and/or `verify_incoming` are set, and depending on whether you're looking at the servers or clients. In general, the cert/key is used to encrypt and/or sign a connection. If `verify_outgoing` is set, then clients will check that the servers they are talking to are using certs to encrypt and sign the connection. If `verify_incoming` is set, then servers will check that clients are using certs to encrypt and sign the connection.

brikis98: I think the technical answer is they can be used for both encryption and verification, but using the lingo from the official docs is fine.

modules/install-consul/install-consul
```diff
@@ -212,10 +216,31 @@ function install_binaries {
...
+
+  local readonly consul_tls_certs_path="$path/tls/"
+  local readonly ca_certs_path="$consul_tls_certs_path/ca/"
+
+  sudo cp "$ca_file_path" "$ca_certs_path"
+  sudo cp "$cert_file_path" "$consul_tls_certs_path"
+  sudo cp "$key_file_path" "$consul_tls_certs_path"
+
+  sudo chown -R "$user:$user" "$consul_tls_certs_path"
+  sudo chmod -R u=rX,g=,o= "$consul_tls_certs_path"
```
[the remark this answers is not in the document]
patoarvizu: To set execute permissions on the `tls/ca` directory, otherwise `-R` would only set `u=r` and botch the directory. I could do pretty much the same with `find`/`-exec` if you think that's cleaner:
```
sudo find "$consul_tls_certs_path/" -type f -exec chmod u=r,g=,o= {} \;
```

modules/run-consul/run-consul
```diff
@@ -227,6 +267,9 @@ EOF
   "ui": $ui
 }
 EOF
+)
+  log_info "Installing Consul config file in $config_path"
+  echo "$default_config_json" "$gossip_encryption_json" "$rpc_encryption_json" | jq --slurp add > "$config_path"
```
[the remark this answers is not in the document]
patoarvizu: There is a reason but it has more to do with just visual alignment and general OCD-ness.

If I set for example `rpc_encryption_json` the same way `retry_join_json` is set, then either the first line of the `HEREDOC` would have to be misaligned:
```
    rpc_encryption_json=$(cat <<EOF
"verify_outgoing": true,
  "verify_incoming": true,
  "ca_path": "$ca_path",
  "cert_file": "$cert_file_path",
  "key_file": "$key_file_path"
EOF
)
```
Or the `HEREDOC` of `default.json` would have a misaligned variable, or the resulting `default.json` would have misaligned fields.

How about I switch it back to setting `gossip_encryption_json` and `rpc_encryption_json` the same way `bootstrap_expect` and `retry_join_json` are set, **but** I keep all the configuration in a single variable holding a json document and pass it through `jq` (just `jq '.'`) to clean it up right before writing to `default.json`?

modules/install-consul/install-consul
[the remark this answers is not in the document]
brikis98: Oh, I see, you're modifying the dir. Yea, it's probably better to modify the files within the dir to all be solely read-only for the owner. Using `find` seems like a reasonable approach. Alternatively, just call `chmod` three times, once on each file path, just as the three calls to `cp`.

modules/run-consul/run-consul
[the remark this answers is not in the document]
brikis98: Ah, gotcha. Yea, passing through `jq` sounds great. It'll catch invalid JSON too.

patoarvizu:
Hi! Just checking... any other issues I should work on here?

I forgot to mention I tested running the example using both images created by Packer (Ubuntu and Amazon Linux), both with and without encryption and it worked fine.

brikis98:
https://github.com/hashicorp/terraform-aws-consul/releases/tag/v0.3.2

PR #61, opened 2018-04-23
Install Git on Amazon Linux

rszalski:
Amazon Linux does not have Git installed by default. The first provisioner step makes sure we install it. `|| true` is used for compatibility with Ubuntu-based builder which does not have `yum` package manager.

Otherwise Amazon Linux will fail with:

    ==> amazon-linux-ami: Provisioning with shell script: /var/folders/zr/k8_fxx5d1xx56n1m758byg4r0000gn/T/packer-shell218591116
    amazon-linux-ami: /tmp/script_3975.sh: line 2: git: command not found
    ==> amazon-linux-ami: Terminating the source AWS instance...

### DIFF
--- a/examples/consul-ami/README.md
+++ b/examples/consul-ami/README.md
@@ -72,6 +72,7 @@ Your code should look more like this:
   "provisioners": [{
     "type": "shell",
     "inline": [
+      "sudo yum install -y git || true",
       "git clone --branch <MODULE_VERSION> https://github.com/hashicorp/terraform-aws-consul.git /tmp/terraform-aws-consul",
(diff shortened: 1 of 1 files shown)


examples/consul-ami/README.md
```diff
@@ -72,6 +72,7 @@ Your code should look more like this:
   "provisioners": [{
     "type": "shell",
     "inline": [
+      "sudo yum install -y git || true",
```
brikis98: Ah, good fix. I'm not a fan of the `|| true`, as Ubuntu users will get a confusing error message.

How about just putting a `NOTE` just below this sample code that says Amazon Linux users will need to install Git?

PR #62, opened 2018-04-23
Ability to specify additional security groups

icy-arctic-fox:
This PR adds the ability to specify additional security groups to attach to the launch configuration. An example usage is a "remote access" security group. The security group lists the IP addresses and systems that can remotely access the system directly. That contains multiple rules that are easier to manage as a separate group, instead of providing the rules directly.

A new, optional variable has been added: `additional_security_group_ids`
This variable accepts a list of security group IDs to attach to the launch configuration. By default, it is empty.

**Downtime needed:** No.
This change does not require downtime. The default behavior does not force a change.

**Backwards compatible:** Yes.
The default behavior works identically as it does without this PR.

The change was tested in our AWS account in three scenarios:
1. Existing cluster without specifying the `additional_security_group_ids` variable. No changes showed up in the plan.
2. Existing cluster with the `additional_security_group_ids` variable. A new launch configuration was created and the ASG was updated to use it. No instances were stopped or created.
3. New cluster with the `additional_security_group_ids` variable. Attached all security groups as expected.


brikis98:
Oh man, I was browsing PRs, and realized that I somehow missed this one. My apologies. No idea how that happened 😞 

The code change looks good, and this is indeed a handy feature, so thank you for the PR. Not sure if you're still at all interested, but the only thing I'd want to ask is how you tested the change?

brikis98:
Understood. Looks like this is conflicting with something recently merged to master. Could you resolve those conflicts?

brikis98:
Thank you! Merging now. Once tests pass, I'll issue a new release.

brikis98:
https://github.com/hashicorp/terraform-aws-consul/releases/tag/v0.3.8

PR #75, opened 2018-08-21
Add circle 2.0 support

Etiene:
Adds support to circle ci 2


.circleci/config.yml
```diff
@@ -0,0 +1,38 @@
...
+          key: external-dep
+          paths:
+            - $HOME/terraform
+            - $HOME/packer
+      - save_cache:
+          key: dep-{{ checksum "test/Gopkg.lock" }}
+          paths:
+            - $HOME/.go_workspace/src/github.com/hashicorp/terraform-aws-consul/test/vendor
+      - run:  if [[ $CIRCLE_BRANCH != "publish-amis" ]]; then run-go-tests --path test; fi
+      - deploy:
```
Etiene: Workflows would be ideal to control the flow regarding when or not to test or publish the amis. However because circle 2 runs each job in independent docker images, that would require either 

1. A complex sharing of the environments. Which is not possible because lots of the dependencies get installed at hardcoded directories like `usr/local/bin`
2. Gruntwork's premade docker image with the dependencies installed, which this repo does not have access because it is in a different github org.

So I chose to keep a good ol' `if`

brikis98: What does the GitHub org have to do with accessing a Docker image? I believe Docker auth is what we actually need here, and we have the proper AWS creds for ECR access. You just need to login...

Etiene: ^ @mcalhoun

Etiene:
TODO: Rebase & update ami publishing given #77

brikis98:
#77 is now merged

.circleci/config.yml
```diff
@@ -0,0 +1,69 @@
+workspace_root: &workspace_root
+  /go/src/github.com/hashicorp/terraform-aws-consul
+
+defaults: &defaults
+  working_directory: *workspace_root
+  docker:
+  - image: 087285199408.dkr.ecr.us-east-1.amazonaws.com/circle-ci-test-image-base:latest
```
robmorgan: Just out of interest who owns Account Id `087285199408 `? Do you want to add a comment so future contributors know its a Gruntwork internal account?

brikis98: That's a good idea. @Etiene Could you add a comment explaining this is a Gruntwork Docker image that has `aws`, `jq`, etc installed?

Issue #76, opened 2018-08-24
Latest public AMIs not available (Consul Connect)


awesome:
Was making initial experiments with this module and got an AMI-not-found error by setting the var `ami_id = "ami-0cefe1c6ca6cb38f6"`. Double checked via AWS web-console for `us-east-1` and `us-east-2` and did not find public AMIs referenced in `_docs`: 
https://github.com/hashicorp/terraform-aws-consul/blame/v0.3.8/_docs/amazon-linux-ami-list.md#L19

Via AWS web-console looked up owner account `562637147889` filtered to 2018 which resulted with no public Consul AMIs… 
https://github.com/hashicorp/terraform-aws-consul/blame/v0.3.8/main.tf#L33

This is probably a backlog issue, but thought mentioning it would be helpful for those wanting to easily try Consul Connect or new UI without using Packer to build AMIs.

```
Error: Error applying plan:

2 error(s) occurred:

* module.consul.module.consul_servers.aws_launch_configuration.launch_configuration: 1 error(s) occurred:

* aws_launch_configuration.launch_configuration: Expected to find a Root Device name for AMI (ami-0cefe1c6ca6cb38f6), but got none
* module.consul.module.consul_clients.aws_launch_configuration.launch_configuration: 1 error(s) occurred:

* aws_launch_configuration.launch_configuration: Expected to find a Root Device name for AMI (ami-0cefe1c6ca6cb38f6), but got none
```

brikis98:
Thanks for the report. Something has indeed broken with our AMI publishing code. I'll look into it.

PR #77, opened 2018-08-24
Change how we publish public AMIs

brikis98:
Our AMI publishing script seemed to be broken (https://github.com/hashicorp/terraform-aws-consul/issues/76). I was struggling to figure out why and realized that Packer can now do most of what that script does with the `ami_regions` and `ami_groups` parameters. I made those changes... and then realized that the actual issue was that we were publishing the AMIs to the wrong account (to Phx DevOps rather than our public AMIs account). Doh.

Overall, I think this solution is simpler and better overall, so perhaps my work has not gone to waste.


_ci/publish-amis.sh
```diff
@@ -7,51 +7,35 @@
...
+echo "Building Packer template $packer_template_path (builder: $builder_name) and sharing it with all AWS accounts in the following regions: $all_aws_regions"
+
+# Copying AMIs to many regions can take longer than Packer's default wait timeouts, so we increase them here per
+# https://github.com/hashicorp/packer/issues/6536
+export AWS_MAX_ATTEMPTS=240
+export AWS_POLL_DELAY_SECONDS=15
+
+# We publish the AMIs to a different AWS account, so set those credentials
+export AWS_ACCESS_KEY_ID="$PUBLISH_AMI_AWS_ACCESS_KEY_ID"
+export AWS_SECRET_ACCESS_KEY="$PUBLISH_AMI_AWS_SECRET_ACCESS_KEY"
```
brikis98: The two lines above are the actual bug fix. The rest is just simplifying how we publish AMIs.

circle.yml
```diff
@@ -27,13 +27,5 @@ deployment:
...
-    branch: publish-amis
-    commands:
-      # We generally only want to build AMIs on new releases, but when we are setting up AMIs in a new account for the
-      # first time, we want to build the AMIs but NOT run automated tests, since those tests will fail without an existing
-      # AMI already in the AWS Account.
-      - ~/$CIRCLE_PROJECT_REPONAME/_ci/publish-amis.sh "ubuntu16-ami"
-      - ~/$CIRCLE_PROJECT_REPONAME/_ci/publish-amis.sh "amazon-linux-ami"
\ No newline at end of file
+      - ~/$CIRCLE_PROJECT_REPONAME/_ci/publish-amis.sh "/home/ubuntu/$CIRCLE_PROJECT_REPONAME/examples/consul-ami/consul.json" "ubuntu16-ami"
+      - ~/$CIRCLE_PROJECT_REPONAME/_ci/publish-amis.sh "/home/ubuntu/$CIRCLE_PROJECT_REPONAME/examples/consul-ami/consul.json" "amazon-linux-ami"
```
Etiene: I'll have to update the circle PR with those 👍

examples/consul-ami/consul.json
```diff
@@ -2,7 +2,9 @@
   "min_packer_version": "0.12.0",
   "variables": {
     "aws_region": "us-east-1",
-    "consul_version": "1.2.2",
+    "consul_version": "1.0.0",
```
Etiene: You dowgraded the default version?

circle.yml
brikis98: Yes plz!

examples/consul-ami/consul.json
brikis98: Whoops. Unintentional, will fix. Good catch 👍

brikis98:
Test failure is due to lack of a published AMI! Merging now and will create a new release to check this all works.

PR #80, opened 2018-08-28
Add test for consul enterprise installation

Etiene:
* Adds test for consul enterprise installation
* Fixes `download-url` var check on ubuntu

@infosecgithub

### DIFF
--- a/examples/consul-ami/consul.json
+++ b/examples/consul-ami/consul.json
@@ -3,7 +3,7 @@
   "variables": {
     "aws_region": "us-east-1",
     "consul_version": "1.2.2",
-    "download_url": ""
+    "download_url": "{{env `CONSUL_DOWNLOAD_URL`}}"
   },
   "builders": [{
     "name": "ubuntu16-ami",
--- a/examples/consul-ami/consul.json
+++ b/examples/consul-ami/consul.json
@@ -51,7 +51,7 @@
   },{
     "type": "shell",
     "inline": [
-      "if [[ -n '{{user `download_url`}}' ]]; then",
+      "if test -n \"{{user `download_url`}}\"; then",
       " /tmp/terraform-aws-consul/modules/install-consul/install-consul --download-url {{user `download_url`}};",
       "else",
       " /tmp/terraform-aws-consul/modules/install-consul/install-consul --version {{user `consul_version`}};",
--- a/modules/install-consul/install-consul
+++ b/modules/install-consul/install-consul
@@ -205,7 +205,7 @@ function fetch_binary {
     download_url="https://releases.hashicorp.com/consul/${version}/consul_${version}_linux_amd64.zip"
   fi
 
-  log_info "Downloading Consul from $download_url to $DOWNLOAD_PACKAGE_PATH"
+  log_info "Downloading Consul to $DOWNLOAD_PACKAGE_PATH"
   curl -o "$DOWNLOAD_PACKAGE_PATH" "$download_url" --location --silent --fail --show-error
 }
 
--- a/modules/install-consul/install-consul
+++ b/modules/install-consul/install-consul
@@ -337,7 +337,7 @@ function install {
   if command -v consul; then
     log_info "Consul install complete!";
   else
-    log_info "Consul was not installed properly.  Aborting.";
+    log_info "Could not find consul command. Aborting.";
     exit 1;
   fi
 }
--- a/modules/install-consul/install-consul
+++ b/modules/install-consul/install-consul
@@ -334,7 +334,12 @@ function install {
     install_tls_certificates "$path" "$user" "$ca_file_path" "$cert_file_path" "$key_file_path"
   fi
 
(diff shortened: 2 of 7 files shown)


examples/consul-ami/consul.json
```diff
@@ -51,7 +51,7 @@
   },{
     "type": "shell",
     "inline": [
-      "if [[ -n '{{user `download_url`}}' ]]; then",
+      "if test -n \"{{user `download_url`}}\"; then",
```
brikis98: Why the change?

modules/install-consul/install-consul
```diff
@@ -334,7 +334,12 @@ function install {
     install_tls_certificates "$path" "$user" "$ca_file_path" "$cert_file_path" "$key_file_path"
   fi
 
-  log_info "Consul install complete!"
+  if command -v consul; then
+    log_info "Consul install complete!";
+  else
+    log_info "Consul was not installed properly.  Aborting.";
```
brikis98: Add "Could not find consul command" to make it clear what about the install failed.

test/consul_helpers.go
```diff
@@ -79,15 +90,20 @@ func runConsulClusterTest(t *testing.T, packerBuildName string, examplesFolder s
...
-	test_structure.RunTestStage(t,"validate", func() {
+	test_structure.RunTestStage(t, "validate", func() {
 		awsRegion := test_structure.LoadString(t, exampleFolder, SAVED_AWS_REGION)
 		terraformOptions := test_structure.LoadTerraformOptions(t, exampleFolder)
+		keyPair := test_structure.LoadEc2KeyPair(t, exampleFolder)
 
-		// Check the Consul servers
-		checkConsulClusterIsWorking(t, CONSUL_CLUSTER_EXAMPLE_OUTPUT_SERVER_ASG_NAME, terraformOptions, awsRegion)
+		if len(enterpriseUrl) > 0 {
+			checkEnterpriseInstall(t, CONSUL_CLUSTER_EXAMPLE_OUTPUT_SERVER_ASG_NAME, terraformOptions, awsRegion, sshUser, keyPair)
```
brikis98: Shouldn't the enterprise install *also* check the cluster is working?

modules/install-consul/install-consul
brikis98: One other thought: check if the `install-consul` script logs the download URL anywhere...

test/consul_helpers.go
Etiene: done

examples/consul-ami/consul.json
brikis98: That... sounds very odd.

BTW, as I mentioned in my other PR, it might be worth allowing both `--download-url` and `--version` to be specified, but if the option is set to an empty string, to just ignore it. That would remove the need for an if-statement here.

modules/install-consul/install-consul
```diff
@@ -205,7 +205,7 @@ function fetch_binary {
     download_url="https://releases.hashicorp.com/consul/${version}/consul_${version}_linux_amd64.zip"
   fi
 
-  log_info "Downloading Consul from $download_url to $DOWNLOAD_PACKAGE_PATH"
+  log_info "Downloading Consul to $DOWNLOAD_PACKAGE_PATH"
```
brikis98: 👍

test/consul_enterprise_test.go
```diff
@@ -0,0 +1,17 @@
...
+
+import (
+	"os"
+	"testing"
+)
+
+// Test the example in the root folder
+func TestConsulInstallFromURLWithUbuntuAmi(t *testing.T) {
+	t.Parallel()
+	runConsulClusterTest(t, "ubuntu16-ami", ".", "../examples/consul-ami/consul.json", "ubuntu", os.Getenv("CONSUL_AMI_TEMPLATE_VAR_DOWNLOAD_URL"))
```
brikis98: See my comment in the Vault PR about adding a separate method that reports a clear error if this env var is not set and has a comment explaining why we use an env var.

examples/consul-ami/consul.json
Etiene: Oh that was the original plan in fact, but after discussing with Josh we came to the conclusion that this makes it clearear to users that these two settings are independent and one of them is required

brikis98: Right, I still think that would be the case. If you specify:

```
--download-url "foo" --version "bar"
```

You'd get an error. But if you specify:

```
--download-url "foo" --version ""
```

That is, an empty string for one of the params, it's ignored, and the other is used. 

And if you specify:

```
--download-url "" --version ""
```

The defaults are used.

Etiene: But the defaults are specified in the packer template, not in the install script.

Etiene: You can already pass both of them to the script, it is not clear from the template which one takes precedence. This is why this if is there. To make it clear.

Issue #42, opened 2018-01-30
Enhancement: Add iam_role_id as variable in root main.tf to permit referencing pre-built IAM roles instead of creating them


kmcquade:
Right now, if you run the terraform-aws-vault role as a user who does not have permissions to create a role (but can assign them), then this build will not be successful due to the need to create a role that just grants describe-instances, describe-groups, and describe-tags. I'd like to be able to supply the iam_role_id, but that's not possible. 

https://github.com/hashicorp/terraform-aws-consul/blob/master/modules/consul-iam-policies/main.tf

I will submit a PR later to implement this. Figured I would create the issue and reference it after.

shdobxr:
+1

thenom:
I am currently working on version that would do this but am still testing:

https://github.com/thenom/terraform-aws-consul

eedwards-sk:
Any updates on this? This is a huge gap for my use case.

I need special permissions to allow my instance to grab certificates at boot, which as it stands, cannot be modified with the current architecture.

Ideally, at least please let me specify additional policy rules, or override the module used.

PR #81, opened 2018-09-10
Switchable IAM

thenom:
Please review my PR that allows the IAM role and policy creation to be disabled.  This allows the role to be created externally and the ARN of the role to be supplied via tf variable.

This helps in situations where end users roles might not have the right to modify IAM but have the rights to deploy standard compute resources.

ref:
https://github.com/hashicorp/terraform-aws-consul/issues/42

I am afraid i am not very good with go so have not updated the tests but i am not in a situation currently where i can test with IAM enabled anyway so if someone could please confirm that all is good with this and merge it in, if not then please let me know.

### DIFF
--- a/modules/consul-cluster/main.tf
+++ b/modules/consul-cluster/main.tf
@@ -186,8 +186,6 @@ resource "aws_iam_role" "instance_role" {
 }
 
 data "aws_iam_policy_document" "instance_role" {
-  count = "${var.enable_iam_setup}"
-
   statement {
     effect  = "Allow"
     actions = ["sts:AssumeRole"]
--- a/modules/consul-cluster/main.tf
+++ b/modules/consul-cluster/main.tf
@@ -161,7 +161,7 @@ resource "aws_iam_instance_profile" "instance_profile" {
 
   name_prefix = "${var.cluster_name}"
   path        = "${var.instance_profile_path}"
-  role        = "${aws_iam_role.instance_role.name}"
+  role        = "${element(coalescelist(aws_iam_role.instance_role.*.name,list("")),0)}"
 
   # aws_launch_configuration.launch_configuration in this module sets create_before_destroy to true, which means
   # everything it depends on, including this resource, must set it as well, or you'll get cyclic dependency errors
--- a/modules/consul-cluster/main.tf
+++ b/modules/consul-cluster/main.tf
@@ -39,7 +39,7 @@ resource "aws_autoscaling_group" "autoscaling_group" {
       value               = "${var.cluster_tag_value}"
       propagate_at_launch = true
     },
-    "${var.tags}"
+    "${var.tags}",
   ]
 }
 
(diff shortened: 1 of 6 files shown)


modules/consul-cluster/main.tf
```diff
@@ -39,7 +39,7 @@ resource "aws_autoscaling_group" "autoscaling_group" {
       value               = "${var.cluster_tag_value}"
       propagate_at_launch = true
     },
-    "${var.tags}"
+    "${var.tags}",
```
brikis98: Extra comma?

modules/consul-cluster/main.tf
```diff
@@ -170,6 +172,8 @@ resource "aws_iam_instance_profile" "instance_profile" {
 }
 
 resource "aws_iam_role" "instance_role" {
+  count = "${var.enable_iam_setup}"
+
   name_prefix        = "${var.cluster_name}"
   assume_role_policy = "${data.aws_iam_policy_document.instance_role.json}"
```
brikis98: This should now be `element(concat(data.aws_iam_policy_document.instance_role.*.json, list("")), 0)`

modules/consul-cluster/variables.tf
```diff
@@ -195,3 +195,13 @@ variable "tags" {
   type        = "list"
   default     = []
 }
+
+variable "enable_iam_setup" {
+  description = "If false, the IAM policy and role setup will be disabled.  This is to allow the IAM to be setup elsewhere if required."
```
brikis98: A bit awkward to read the negative first. Perhaps, "If true, create the IAM Role, IAM Instance Profile, and IAM Policies. If false, these will not be created, and you can pass in your own IAM Instance Profile via var.iam_instance_profile_name".

modules/consul-iam-policies/variables.tf
```diff
@@ -6,3 +6,8 @@
 variable "iam_role_id" {
   description = "The ID of the IAM Role to which these IAM policies should be attached"
 }
+
+variable "enabled" { 
+  description = "Give the option to disable this module if required"
+  default = true
```
brikis98: Please run `terraform fmt` on all of the code.

modules/consul-cluster/main.tf
thenom: Ah yes, this would have come up if i could test the IAM enabled mode.  I will correct this.

modules/consul-cluster/variables.tf
thenom: No problem i will replace.

modules/consul-iam-policies/variables.tf
thenom: Usually i have pre-commit to do this.  I will add the pre-commit.yml to do this on commit if you have no issue with it?  In the meantime i will run fmt on it.

thenom:
In our infrastructure we have separate roles, one for the IAM control and one for BAU access which is the one we use for terraform deploys.  This is why i cannot test the IAM enabled process but with it set to false my consul cluster fully deploys without issues when i provide the role i created using the IAM role.

modules/consul-cluster/main.tf
thenom: I have just removed the count from the data resource as that was un-required due to it not needing any IAM rights to do so.  That makes this change un-required also.

thenom:
Ok, i submitted a new commit.  Let me know if there are any issues.  I have redeployed the cluster again with these changes and it still works fine.

modules/consul-cluster/main.tf
```diff
@@ -157,6 +157,8 @@ module "security_group_rules" {
 # ---------------------------------------------------------------------------------------------------------------------
 
 resource "aws_iam_instance_profile" "instance_profile" {
+  count = "${var.enable_iam_setup}"
+
   name_prefix = "${var.cluster_name}"
   path        = "${var.instance_profile_path}"
   role        = "${aws_iam_role.instance_role.name}"
```
brikis98: This needs the `element(coalescelist(...))` workaround, as `aws_iam_role.instance_role` is now a list

thenom: Ah yes, well spotted, changing now.

PR #93, opened 2018-10-26
adds autoscaling group id and arn

eedwards-sk:
I cannot create a load balancer for this module's ASG without additional outputs.


eedwards-sk:
Closing. Got it working with just the name :)

PR #94, opened 2018-10-26
Add variable for ASG service role

briandbecker:
* When not specified; the created ASG will use the default AWS Service
  Role for Auto Scaling for performing AWS API calls such as
  RunInstances.
* This change allows the caller to specify a custom ASG role by passing
  role ARN and as such allowing custom permissions such as access to a
  KMS key for encrypted AMIs
* By defaulting to empty the aws provider will use the default role

PR #95, opened 2018-11-01
Fix broken test for custom ASG service-linked role

briandbecker:
* Test did not account for the role already existing
* Adds timestamp to role name to make unique
* Due to limits in the AWS role name length the uuid function won't work
* Due to limits in the AWS role name supported characters it has to do
  some funky replace statements on the timestamp

### DIFF
--- a/examples/example-with-custom-asg-role/main.tf
+++ b/examples/example-with-custom-asg-role/main.tf
@@ -17,7 +17,7 @@ terraform {
 # ---------------------------------------------------------------------------------------------------------------------
 resource "aws_iam_service_linked_role" "consul_asg_role" {
   aws_service_name = "autoscaling.amazonaws.com"
-  custom_suffix    = "${format("consul-slr-%s",replace(replace(timestamp(),"-",""),":",""))}"
+  custom_suffix    = "${var.consul_service_linked_role_suffix}"
   description      = "Service-Linked Role enables access to AWS Services and Resources used or managed by Auto Scaling"
 }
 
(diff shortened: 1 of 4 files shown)


briandbecker:
This should fix the issue here https://github.com/hashicorp/terraform-aws-consul/pull/94

examples/example-with-custom-asg-role/main.tf
```diff
@@ -17,7 +17,7 @@ terraform {
 # ---------------------------------------------------------------------------------------------------------------------
 resource "aws_iam_service_linked_role" "consul_asg_role" {
   aws_service_name = "autoscaling.amazonaws.com"
-  custom_suffix    = "test-consul-service-linked-role"
+  custom_suffix    = "${format("consul-slr-%s",replace(replace(timestamp(),"-",""),":",""))}"
```
brikis98: What about adding a `custom_suffix` variable to this example set to something simple by default (e.g., `test-consul-service-linked-role` as you had before) and then overriding it at test time with some name that includes the `UniqueId()` from Terratest?

briandbecker: I'm new to terratest so will have to figure that one out; but sounds like a better fix overall...

briandbecker:
@brikis98 gave it a shot and even tried to keep it DRY.  Looking forward to your thoughts.

brikis98:
OK, tests passed! New release is here: https://github.com/hashicorp/terraform-aws-consul/releases/tag/v0.4.1

PR #97, opened 2018-11-09
Fix: Retry downloading consul on install script

Etiene:
Had an occasional failure on vault test suite because consul failed to download.

### DIFF
--- a/modules/install-consul/install-consul
+++ b/modules/install-consul/install-consul
@@ -41,30 +41,30 @@ function print_usage {
 }
 
 function log {
-  local readonly level="$1"
-  local readonly message="$2"
-  local readonly timestamp=$(date +"%Y-%m-%d %H:%M:%S")
+  local -r level="$1"
+  local -r message="$2"
+  local -r timestamp=$(date +"%Y-%m-%d %H:%M:%S")
   >&2 echo -e "${timestamp} [${level}] [$SCRIPT_NAME] ${message}"
 }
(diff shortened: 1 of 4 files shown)


modules/install-consul/install-consul
```diff
@@ -86,6 +86,30 @@ function assert_either_or {
   fi
 }
 
+# A retry function that attempts to run a command a number of times and returns the output
+function retry {
+  local readonly cmd=$1
+  local readonly description=$2
```
brikis98: `local -r`

brikis98: Also, wrap arguments in quotes

modules/run-consul/run-consul
```diff
@@ -316,14 +316,14 @@ EOF
...
-  local readonly consul_user="$6"
+  local -r supervisor_config_path="$1"
+  local -r consul_config_dir="$2"
+  local -r consul_data_dir="$3"
+  local -r consul_log_dir="$4"
+  local -r consul_bin_dir="$5"
+  local -r consul_user="$6"
   shift 6
-  local readonly environment=("$@")
+  local -r environment=("$@")
```
brikis98: There's one gotcha here: all arrays should use `-ar`. Otherwise, they don't seem to be properly locally scoped in some versions of bash. We had a long discussion in #troubleshooting about this and it took a while to figure that out!

PR #110, opened 2019-01-21
Fix switchable iam bug

Etiene:


modules/consul-cluster/main.tf
```diff
@@ -57,7 +57,7 @@ resource "aws_launch_configuration" "launch_configuration" {
   user_data     = "${var.user_data}"
   spot_price    = "${var.spot_price}"
 
-  iam_instance_profile        = "${element(coalescelist(list(var.iam_instance_profile_name),aws_iam_instance_profile.instance_profile.*.name),0)}"
+  iam_instance_profile        = "${var.enable_iam_setup ? element(aws_iam_instance_profile.instance_profile.*.name,0) : var.iam_instance_profile_name}"
```
brikis98: Terraform's conditional syntax is NOT short circuiting. That means the `element(aws_iam_instance_profile.instance_profile.*.name,0)` portion will be evaluated even if `enable_iam_setup` is false and that resource is not created, which will lead to the `element` call throwing an exception for trying to look up an item in an empty list. So you'll need the `element(concat(..., list("")), 0)` workaround here.

PR #114, opened 2019-02-14
add support to variable "subnet_ids"

davidbennatan:
add support to variable "subnet_ids"


davidbennatan:
conditional operator cannot be used with list values

PR #115, opened 2019-02-14
add subnet ids variables

davidbennatan:
add subnet ids variables

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -71,7 +71,7 @@ module "consul_servers" {
   user_data = "${data.template_file.user_data_server.rendered}"
 
   vpc_id     = "${data.aws_vpc.default.id}"
-  subnet_ids = "${length(var.subnet_ids) == 0 ? data.aws_subnet_ids.default.ids : var.subnet_ids}" 
+  subnet_ids = "${split(",",join(",",var.subnet_ids) == ""  ? join(",",data.aws_subnet_ids.default.ids) :join(",",var.subnet_ids))}" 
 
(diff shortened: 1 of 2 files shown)


davidbennatan:
I have a vpc with both private and public networks, so i need the option to chose the elastic group subnet location.
I test on my private enviroment on my fork, with both inserted and default  arguments

brikis98:
OK, important sanity check: [as mentioned here](https://github.com/hashicorp/terraform-aws-consul#how-to-use-this-module), [here](https://github.com/hashicorp/terraform-aws-consul/tree/master/examples/root-example), and [here](https://github.com/hashicorp/terraform-aws-consul/blob/master/main.tf#L3) the code in the root of this module is *example code*. It is meant to show one simple way to use the modules in the `modules` folder and not to be used directly in prod. I appreciate the PR, and the changes you propose are reasonable, but they sound like you are trying to use the code in the root directly for a real use-case...

davidbennatan:
You right, my environment is not yet prod, but i should switch to use the inner modules directly.
nonetheless  I think that the changes will slightly increase the example code flexibility

brikis98:
> I think that the changes will slightly increase the example code flexibility

True, but at the cost of making the examples harder to read, maintain, and test. If you think this is a critical change to have, I can merge it in, but otherwise, I'm not sure it's worth the extra overhead.

davidbennatan:
Well, I cant really call this change critical, I should just switch and use the inner modules directly as I supposed to.
thanks for the review.

PR #118, opened 2019-02-25
Update docs to clarify role of root of repo

brikis98:
We've had a lot of confusion about the purpose of the code in the root of this repo. This is an attempt to improve the docs to make this more clear. Feedback welcome on how to make it even clearer.


devvesa:
One of the confused here. I'd say the main confusion comes the fact that there is already an `examples` directory on the main root, so one may think that the `main.tf` of the root directory is also production ready.

brikis98:
Completely agree it's confusing, but for better or worse, it's a hard requirement of the Terraform Registry 😕

PR #123, opened 2019-03-08
update default num servers and clients

iainthegray:
This is a PR to update the default number of consul servers to 5 as per the hashicorp Reference Architecture and to set the default number of clients to 0

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -24,12 +24,12 @@ variable "cluster_name" {
 
 variable "num_servers" {
   description = "The number of Consul server nodes to deploy. We strongly recommend using 3 or 5."
-  default     = 3
+  default     = 5
 }
 
 variable "num_clients" {
   description = "The number of Consul client nodes to deploy. You typically run the Consul client alongside your apps, so set this value to however many Instances make sense for your app code."
-  default     = 6
+  default     = 0
 }
 
 variable "cluster_tag_key" {


iainthegray:
> Quick sanity check: are you aiming for the root example to implement the Reference Architecture recommendations?
Yes. 
>If so, perhaps the better option is to just remove the consul clients from the root example entirely and only have them in other examples?
I am unsure what the consul-clients do in infrastructure terms. Sure you need a consul client on something, but I cant see a usecase where you would just build an instance with a consul client on it. Hence that is why I have set the default consul clients to 0 then _if_ people want one then they can change that variable

brikis98:
> I am unsure what the consul-clients do in infrastructure terms. Sure you need a consul client on something, but I cant see a usecase where you would just build an instance with a consul client on it. Hence that is why I have set the default consul clients to 0 then if people want one then they can change that variable

The `consul-clients` accomplish two things:

1. They show users an example of how to install the consul agent and run it in consul mode. This pattern can be copied to all your apps/services that need to talk to Consul servers.
1. Since we have automated tests for all of our example code, including the root example, having the consul clients here provides a way to test that the code in this repo works correctly in client mode.

If the goal with the root example is to implement a Consul Ref Arch, then there's no need to have the client code here. Instead, you could remove it from here, and just ensure one of the other examples has it.

iainthegray:
OK so I understand that, but this PR is simply changing default values in the module. The function of a default value is to show a sensible value that most people can leave untouched. Currently the default for the consul clients is 6 which means that _everyone_ except your tests will have to amend it. 
Hence I believe it should be set to 0

brikis98:
> Currently the default for the consul clients is 6 which means that everyone except your tests will have to amend it. Hence I believe it should be set to 0

Any reason you don't want to just remove the code for the clients *entirely( from the root example if the plan is to make it the RA?

PR #124, opened 2019-03-08
Fix ami publishing and add nightly builds

Etiene:
It works as seen on https://circleci.com/gh/hashicorp/terraform-aws-consul/tree/publish-amis

Additionally, "gruntwork-ci" github user needed access to this repo so it could publish the markdown with the resulting amis

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -50,9 +50,10 @@ jobs:
     steps:
       - checkout
       - run: echo 'export PATH=$HOME/terraform:$HOME/packer:$PATH' >> $BASH_ENV
-      - run: gruntwork-install --module-name "aws-helpers" --repo "https://github.com/gruntwork-io/module-ci" --tag "v0.12.2"
-      - run: gruntwork-install --module-name "git-helpers" --repo "https://github.com/gruntwork-io/module-ci" --tag "v0.12.2"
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -63,18 +64,31 @@ workflows:
...
+            tags:
+              only: /^v.*/
+  nightly-test:
+    triggers:
+      - schedule:
+          cron: "0 0 * * *"
+          filters:
+            branches:
+              only:
+                - publish-amis
```
josh-padnick: Should this be `ignore` and not `only`? Either way, it'd be helpful to add a comment here on what the nightly test is actually testing: All tests or just the publish-ami functionality.

josh-padnick:
Thanks for submitting this fix! This PR suggests the only change was the need to add `sudo` on the `gruntwork-install` commands but I think I saw in Slack the real culprit was that our machine key stopped working? Either way, could you update this PR's text with the relevant conclusions? Otherwise, LGTM!

.circleci/config.yml
Etiene: Oh it should be only: - master. Forgot to switch back after I tested.

josh-padnick:
@Etiene Do you need to issue a new release to build new AMIs on this?

PR #128, opened 2019-03-26
Add latest AMI IDs. [ci skip]

Etiene:


josh-padnick:
@Etiene I noticed that (1) the Amazon Linux AMIs weren't updated, and (2) that the CI tool committed the Ubuntu updates to a branch. Does it make sense to update the Circle CI script to include the Amazon Linux AMIs and push to master so no PR is required?

PR #130, opened 2019-03-28
Switch from supervisord to systemd

### LINKED ISSUE #86: enhancement request: amazon linux 2 support
[amazon linux 2](https://aws.amazon.com/amazon-linux-2/) has a suite of numerous updates and improvements

however, with the bump in centos it's based on, it now uses systemd instead of supervisord, so I suspect the install script will need testing and additional environment checks to use the appropriate installation method

edtan:
Closes #86 by switching to Amazon Linux 2 and changing to `systemd`.

The `consul.service` configuration was based off of the configuration listed at the [Consul Deployment Guide](https://learn.hashicorp.com/consul/advanced/day-1-operations/deployment-guide#configure-systemd).


Etiene:
Thanks for the PR! I'm gonna run the test suite now

modules/run-consul/run-consul
```diff
@@ -315,37 +316,45 @@ EOF
...
+  local -r config_path="$consul_config_dir/$CONSUL_CONFIG_FILE"
+
+  log_info "Creating systemd config file to run Consul in $systemd_config_path"
+  cat > "$systemd_config_path" <<EOF
+[Unit]
+Description="HashiCorp Consul - A service mesh solution"
+Documentation=https://www.consul.io/
+Requires=network-online.target
+After=network-online.target
+ConditionFileNotEmpty=$config_path
```
brikis98: Same question here about stdout/stderr logs. If we want users to use `journalctl`, we should document how to do so.

edtan: Ah yes, good point.  I've added some documentation about `journalctl` to view the logs, and also made the `StandardOutput` and `StandardError` config options available in case anyone wants to use them.

edtan:
No problem!  Thank you for making these modules publicly available.  They've been super useful!

Glad to hear the tests passed.  I'm still trying to set up the tests on my machine, but have been testing the changes manually in my test environments.

brikis98:
Looks like the last commit had a test failure?

brikis98:
@Etiene Could you take over getting this PR passing tests and merged from here?

edtan:
whoops, sorry!  I'll try to re-run some tests locally this week, but if you have some stack traces and can give me some pointers, I'll be glad to rewrite the tests as necessary.  Note that I don't have access to the enterprise images, and I'm kinda stingy on AWS fees so I try to limit launching instances :) but I kid - I'll run the relevant tests if I know which ones are failing. :))

edtan:
Strange, I re-ran the (non-enterprise) tests again and they've passed now.  Were the tests failing on the enterprise test case?

brikis98:
Looks like the test failure was `TestConsulClusterWithUbuntuAmi` due to: 

```
TestConsulClusterWithUbuntuAmi 2019-03-29T12:31:29Z command.go:94: 1553862689,,ui,error,==> ubuntu16-ami: Error creating AMI: InvalidAMIName.Duplicate: AMI name consul-ubuntu-2019-03-29T12-28-48Z is already in use by AMI ami-0f589d96558469b10\n==> ubuntu16-ami: 	status code: 400%!(PACKER_COMMA) request id: 10801e75-981f-4231-bf0a-84a278a456ae
```

This is probably a minor, intermittent timing issue in the tests, not related to your changes. So I think this is good to merge!

PR #132, opened 2019-04-04
Add iam_role_name to outputs of consul-cluster module. Fixes #106

### LINKED ISSUE #106: Need IAM Policy Name not ARN/ID
Terraform doesn't allow iam role lookups by ID or ARN, so it doesn't help much that you guys return that info, especially given you're doing name_prefix and not name in your call.

tomaspinho:


brikis98:
Is this not the same as https://github.com/hashicorp/terraform-aws-consul/blob/master/modules/consul-cluster/outputs.tf#L21-L24 ?

tomaspinho:
@brikis98, IAM Role IDs are not IAM Role Names. The Terraform AWS Provider takes the latter in most cases as input to other resources.

brikis98:
Are you 100% sure? We use ID all over the place and the [docs](https://www.terraform.io/docs/providers/aws/r/iam_role.html#attributes-reference) say that both return the same thing...

tomaspinho:
@brikis98, yes. If you look at these docs, you'll see that these take the IAM Role Name: https://www.terraform.io/docs/providers/aws/r/iam_role_policy_attachment.html

I had to write a "hack" like this in my code:

```

data "aws_arn" "consul_iam_role" {
  arn = "${module.consul_cluster.iam_role_arn}"
}

resource "aws_iam_role_policy_attachment" "consul" {
  role       = "${replace(data.aws_arn.consul_iam_role.resource, "role/", "")}"
  policy_arn = "arn:aws:iam::aws:policy/service-role/..."
}
```

brikis98:
Yea, it's definitely confusing. Closing now, thanks!

PR #135, opened 2019-05-02
Enable systemd consul service

adriananeci:
To make sure that consul service will automatically start after a server reboot it has to be enabled so that systemd will know to autostart it during boot sequence.


brikis98:
Tests passed. Merging. Thank you!

brikis98:
https://github.com/hashicorp/terraform-aws-consul/releases/tag/v0.6.1

PR #149, opened 2019-08-23
Set defaults for allowed_inbound_cidr_blocks

jinnko:
These variables don't necessarily have to be set to anything and leaving
them empty can work just fine due to the security groups allowing
"self".  As such setting them to an empty list avoids extra rules being
put in place.


Etiene:
Thanks for the PR!

PR #151, opened 2019-09-09
Add support for enabling auto_encrypt on both server and client instances

jinnko:
We get support for enabling auto_encryption at set up time, allowing client instances to get their TLS key/cert pairs from the consul servers.

In order to allow browsers to access the UI we also need to enable the HTTPS endpoint and disable mTLS on the HTTPS endponit, while keeping mTLS enabled for RPC connections.

There are a couple of special cases to be considered (see individual commits for details):

1. Workaround for v1.6.0 (hashicorp/consul#6391)
2. Workaround lack of support for HTTPS (hashicorp/consul#6403)

### DIFF
--- a/modules/run-consul/run-consul
+++ b/modules/run-consul/run-consul
@@ -51,6 +51,7 @@ function print_usage {
   echo -e "  --ca-path\t\tPath to the directory of CA files used to verify outgoing connections. Optional. Must be specified with --enable-rpc-encryption."
   echo -e "  --cert-file-path\tPath to the certificate file used to verify incoming connections. Optional. Must be specified with --enable-rpc-encryption and --key-file-path."
   echo -e "  --key-file-path\tPath to the certificate key used to verify incoming connections. Optional. Must be specified with --enable-rpc-encryption and --cert-file-path."
+  echo -e "  --enable-auto-encryption\t\tEnable auto_encrypt setting for servers and clients."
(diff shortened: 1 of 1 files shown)


brikis98:
@Etiene Could you review this one?

modules/run-consul/run-consul
```diff
@@ -302,13 +304,56 @@ EOF
...
 "verify_incoming": true,
+"verify_incoming_https": false,
+"verify_server_hostname": true,
+"enable_agent_tls_for_checks": true,
 "ca_path": "$ca_path",
 "cert_file": "$cert_file_path",
 "key_file": "$key_file_path",
+"ports": {
+  "https": 8501,
+  "grpc": 8502
```
Etiene: Shouldn't these be configurable?

modules/run-consul/run-consul
```diff
@@ -302,13 +304,56 @@ EOF
...
+"ca_file": "$ca_path",
+"auto_encrypt": {
+    "tls": true
+},
+"ports": {
+    "http": 8500,
+    "https": 8501
+},
+"http_config": {
+    "allow_write_http_from": ["127.0.0.0/8"]
```
Etiene: Should this be configurable too?

Etiene:
Thanks for the PR! How did you test this? I wonder if we should have an automated test to check this behavior

bigeasy:
I created a Pull Request for this Pull request.

https://github.com/ixydo/terraform-aws-consul/pull/1

Enable auto encryption without requiring server key and certificate. My understand of auto-encryption is that the server will provide a PKI certificate to clients, you don't have to distribute them yourself. Adjusted to create a configuration similar to on in the Learning Consul documentation.

PR #159, opened 2019-12-04
Don't install dnsmasq on ubuntu18

yorinasub17:
This fixes the AMI example so that dnsmasq installation step is skipped for Ubuntu 18.


yorinasub17:
FYI I don't have merge permissions on this repo so you will have to do the honors :)

PR #161, opened 2019-12-09
Add encrypted option for root block device and add ebs specifications

szaki99:
The goal is to have encrypted root block device and ebs encrypted block device.

### DIFF
--- a/modules/consul-cluster/main.tf
+++ b/modules/consul-cluster/main.tf
@@ -75,10 +75,19 @@ resource "aws_launch_configuration" "launch_configuration" {
 
   ebs_optimized = var.root_volume_ebs_optimized
 
+  ebs_block_device {
+    device_name           = var.ebs_device_name
+    volume_size           = var.ebs_volume_size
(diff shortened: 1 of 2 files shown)


modules/consul-cluster/main.tf
```diff
@@ -75,10 +75,19 @@ resource "aws_launch_configuration" "launch_configuration" {
 
   ebs_optimized = var.root_volume_ebs_optimized
 
+  ebs_block_device {
+    device_name           = var.ebs_device_name
+    volume_size           = var.ebs_volume_size
+    volume_type           = var.ebs_volume_type
+    delete_on_termination = var.ebs_volume_delete_on_termination
+    encrypted             = var.ebs_volume_encrypted
+  }
```
brikis98: We should not always add an EBS block device. Instead, add a new `ebs_block_devices` input variable, which contains a list of objects, and use that as the value with `for_each`. The default will be an empty list to maintain backwards compatibility, but now you can override the list to add one or more EBS volumes.

PR #168, opened 2020-03-21
listen on 0.0.0.0 instead of

sachnk:
Fix dnsmasq configuration so that it listens on `0.0.0.0` instead of `$consul_ip`, which by default, is `127.0.0.1`.

I believe the intention here is to allow dnsmasq to accept inbound requests on the default interface, not just your local loopback.

This matters because if you run a docker container on a VM you installed dnsmasq on, you won't be able to target the dnsmasq server if it's listening on `127.0.0.1`.

### DIFF
--- a/modules/install-dnsmasq/install-dnsmasq
+++ b/modules/install-dnsmasq/install-dnsmasq
@@ -106,7 +106,7 @@ function write_consul_config {
 # Enable forward lookup of the '$consul_domain' domain:
 server=/${consul_domain}/${consul_ip}#${consul_port}
 
-listen-address=${consul_ip}
+listen-address=0.0.0.0
 bind-interfaces
 EOF
 }


brikis98:
> I believe the intention here is to allow dnsmasq to accept inbound requests on the default interface, not just your local loopback.

Hm, I'm not sure that's really the case... At least, by default, we only want dnsmasq to be usable from the instance itself. The Docker case is an interesting one, but perhaps that should be opt-in?

ryandens:
Something else to consider would be to parameterize the dnsmasq listen address(s) with a sensible default of `127.0.0.1` but allow users of this script to specify a different address via the command line. A common way to interact with consul from inside a docker container is to use a dummy interface on a reserved link-local address such as 169.254.1.1 as described by [this article](https://medium.com/zendesk-engineering/making-docker-and-consul-get-along-5fceda1d52b9). This would also give users like sachnk the flexibility to use `0.0.0.0` if they so wanted without giving up the desired default behavior of only being available from the instance itself. 

If you think this is a sensible approach, I'd be happy to provide a PR to update the script with the parameter for the dnsmasq listen address. If you think this warrants more discussion, we can continue to discuss here/I can open an issue to formalize the feature request.

If anyone is looking for a workaround for this kind of issue, my team and I use this script on our nomad client instances to install dnsmasq. We use packer/ansible to build our machine images, and use the following ansible steps to get the behavior we need in the meantime

```yaml
    - name: Clone terraform-aws-consul
      git:
        repo: 'https://github.com/hashicorp/terraform-aws-consul.git'
        dest: /tmp/terraform-aws-consul
        version: "v{{ consul_module_version }}"
        depth: 1

    - name: Make dnsmasq listen on our dummy interface
      replace:
        path: /tmp/terraform-aws-consul/modules/install-dnsmasq/install-dnsmasq
        regexp: listen-address=\${consul_ip}
        replace: listen-address=169.254.1.1\nlisten-address=127.0.0.1\n

    - name: Install consul for use in client mode
      command: "/tmp/terraform-aws-consul/modules/install-consul/install-consul --version {{ consul_version }}"

    - name: Install dnsmasq to forward appropriate DNS queries to Consul
      command: "/tmp/terraform-aws-consul/modules/install-dnsmasq/install-dnsmasq"
```

brikis98:
@ryandens Making the dnsmasq IP configurable, while keeping the same default as it is now, sounds like a reasonable solution! A PR is very welcome. I'll close this one in the meantime due to inactivity.

PR #172, opened 2020-05-14
Update tests

yorinasub17:
Continuation of https://github.com/hashicorp/terraform-aws-consul/pull/171

- Convert to go modules
- Pull in latest terratest version
- Make corresponding changes to circleci script to support go modules
- Add support for terratest_log_parser
- Update all packer templates to be compatible with `1.5.4+`

Note that tests are still unstable. Different tests are failing each time with a timeout waiting for consul to come up. I don't quite have time to dig in, but it appears to be intermittent.


yorinasub17:
Thanks for review. Going to merge this in now so that all the future changes get this.

Issue #165, opened 2020-02-23
Adding Consul Connect support to run-consul


bwalendz:
Would you guys be open to a PR adding Consul Connect support to the run-consul script? Because right now I would need to go through the hassle of having another config file just to enable Connect.

Everything defined in https://www.consul.io/docs/agent/options.html#connect.

If this is not the intended way, let me know.

brikis98:
Yes please!

PR #173, opened 2020-05-14
Enable Consul Connect on terraform-aws-consul

7hacker:
This PR addresses issue #165 

### Design

We add a flag, '--enable-connect' to the run-consul module that turns on Consul Connect while bootstrapping a new cluster. This flag creates the basic key-value pair in the servers configuration (default.json).

To use in production, we suggest overriding the default configuration as specified in the Readme. This enables the production deployments to:

1. Adopt the best practices in ensuring secure communications, ACL as specified here: https://learn.hashicorp.com/consul/developer-mesh/connect-production

2. Use an alternative CA provider like Vault or apply specific private_key and root_cert values to the default CA provider by Consul, as specified here: https://learn.hashicorp.com/consul/developer-mesh/connect-production

3. Use Envoy as a proxy as specified here: https://www.consul.io/docs/connect/proxies/envoy

### Examples

We also provide examples showing: 
1. How to setup a cluster with Consul Connect
2. Deploying 3 Services with sidecar proxies
3. Creating intentions between these 3 services and disabling communications from one of them to another

### Test

Test include running through the example and verifying:

1. A cluster comes up and the default Consul CA issues a root certificate
2. Communication occurs via sidecar proxies (disabling a sidecar should disable communication)
3. Intentions are respected. Services disallowed to communicate with other services, cannot.


modules/run-consul/README.md
```diff
@@ -80,6 +80,7 @@ The `run-consul` script accepts the following arguments:
 * `ca-file-path` (optional): Path to the CA file used to verify outgoing connections. Must be specified with `enable-rpc-encryption`, `cert-file-path` and `key-file-path`.
 * `cert-file-path` (optional): Path to the certificate file used to verify incoming connections. Must be specified with `enable-rpc-encryption`, `ca-file-path`, and `key-file-path`.
 * `key-file-path` (optional): Path to the certificate key used to verify incoming connections. Must be specified with `enable-rpc-encryption`, `ca-file-path` and `cert-file-path`.
+* `enable-connect` (optional): If this flag is set, turn on Consul Connect when bootstrapping a cluster. To specify your own CA, specify an override config as outlined below.
```
yorinasub17, line 83: This should link to the corresponding section. Also, there should be a link to the docs for configuring the custom CA somewhere in here.

examples/example-with-consul-connect/user-data-client.sh
```diff
@@ -0,0 +1,15 @@
...
+set -e
+
+# Send the log output from this script to user-data.log, syslog, and the console
+# From: https://alestic.com/2010/12/ec2-user-data-output/
+exec > >(tee /var/log/user-data.log|logger -t user-data -s 2>/dev/console) 2>&1
+
+# These variables are passed in via Terraform template interplation
+/opt/consul/bin/run-consul --client --cluster-tag-key "${cluster_tag_key}" --cluster-tag-value "${cluster_tag_value}"
+
+# You could add commands to boot your other apps here
```
yorinasub17, line 15: If you have time, be sure to come back here and setup an actual service and register it to connect!

7hacker: yes, hoping to get to this point with testing!

test/consul_cluster_with_connect_test.go
```diff
@@ -0,0 +1,26 @@
...
+// Test the example in the example-with-consul-connect folder
+func TestConsulClusterWithUbuntu16Ami(t *testing.T) {
+	t.Parallel()
+	runConsulClusterTest(t, "ubuntu16-ami", ".", "../examples/consul-ami/consul.json", "ubuntu", "")
+}*/
+
+// Test the example in the example-with-consul-connect folder
+func TestConsulConnectWithUbuntu18Ami(t *testing.T) {
+	t.Parallel()
+	runConsulClusterTest(t, "ubuntu18-ami", ".", "../examples/consul-ami/consul.json", "ubuntu", "")
```
yorinasub17, line 17: FYI, this isn't using the user-data scripts of the `examples/example-with-consul-connect` example. You need to modify the root example to make the user data script source configurable and pass in the path in the test.

test/consul_connect_tests.go
```diff
@@ -0,0 +1,137 @@
...
+func runConsulConnectTest(t *testing.T, packerBuildName string, examplesFolder string, packerTemplatePath string, sshUser string) {
+	runConsulConnectTestWithVars(t,
+		packerBuildName,
+		examplesFolder,
+		packerTemplatePath,
+		sshUser,
+		map[string]interface{}{})
+}
+
+func runConsulConnectTestWithVars(t *testing.T, packerBuildName string, examplesFolder string, packerTemplatePath string, sshUser string, terraformVarsMerge map[string]interface{}) {
```
yorinasub17, line 34: NIT: you might be working on this, but to avoid duplicating the code consider passing in a boolean param `checkConnect` which is used to enable connect tests.

modules/run-consul/README.md
yorinasub17: Reminder: update docs to indicate this is only for enabling the server side.

examples/example-with-consul-connect/user-data-client.sh
7hacker: This is done

test/consul_cluster_with_connect_test.go
7hacker: Fixed

test/consul_connect_tests.go
7hacker: I thought about this and realized that for future testing we would need more tests on the service layers that may necessitate a separate helper for connect tests. Further, the overlap is much less now, but yes can be refactored some more

modules/run-consul/README.md
7hacker: Updated  to indicate server flag is required as well

7hacker:
This is ready for review for the Trial project day. There are some concrete next steps here, namely:

1. Set up intentions in the consul clients
2. Test that the proxies do route communications and the intentions are respected
3. A better way to register services and sidecar proxies for clients. Right now this is done crudely in user-data-client script as shown in the example, however this is a challenge if there is more than 1 client and you want to control the placement of services and sidecar proxies
4. Integration with Envoy & Vault is an obvious next request from folks who end up using this in production
5. Better documentation on setting things up

modules/run-consul/run-consul
```diff
@@ -530,6 +546,9 @@ function run {
       --enable-gossip-encryption)
         enable_gossip_encryption="true"
         ;;
+      --enable-connect)
+        enable_connect="true"
```
brikis98, line 550: Same here. Declare above as a `local` variable and pass it explicitly to other functions that need it.

examples/example-with-consul-connect/README.md
```diff
@@ -0,0 +1,33 @@
+# Consul Cluster with Connect service mesh
+
+This folder shows an example of Terraform code that uses the [run-consul module](https://github.com/hashicorp/terraform-aws-consul/tree/master/modules/consul-cluster) to deploy
+a [Consul](https://www.consul.io/) cluster in [AWS](https://aws.amazon.com/) with the Consul Connect Service Mesh turned on. The cluster consists of 2 Services with
+side-proxies and upstream dependencies between them
```
brikis98, line 5:
```suggestion
side-proxies and upstream dependencies between them.
```

examples/example-with-consul-connect/main.tf
```diff
@@ -0,0 +1,176 @@
...
+}
+
+# ---------------------------------------------------------------------------------------------------------------------
+# AUTOMATICALLY LOOK UP THE LATEST PRE-BUILT AMI
+# This repo contains a CircleCI job that automatically builds and publishes the latest AMI by building the Packer
+# template at /examples/consul-ami upon every new release. The Terraform data source below automatically looks up the
+# latest AMI so that a simple "terraform apply" will just work without the user needing to manually build an AMI and
+# fill in the right value.
+#
+# !! WARNING !! These exmaple AMIs are meant only convenience when initially testing this repo. Do NOT use these example
```
brikis98, line 24:
```suggestion
# !! WARNING !! These example AMIs are meant only convenience when initially testing this repo. Do NOT use these example
```

Do we have this typo in all our examples?

examples/example-with-consul-connect/user-data-client.sh
```diff
@@ -0,0 +1,58 @@
...
+
+# Register both services foo & bar
+consul services register /opt/consul/config/serv_foo.json
+consul services register /opt/consul/config/serv_bar.json
+
+# Start a proxy sidecar for service foo
+nohup consul connect proxy -sidecar-for foo &>/dev/null &
+
+# Start a proxy sidecar for service bar
+nohup consul connect proxy -sidecar-for bar &>/dev/null &
```
brikis98, line 58: Do we have the services actually try to communicate with each other here? E.g., Send a request from one to the other?

modules/run-consul/run-consul
```diff
@@ -309,6 +313,17 @@ EOF
 )
   fi
 
+  local connect_configuration=""
+  if [[ "$enable_connect" == "true" && "$server" == "true" ]]; then
```
brikis98, line 317: `enable_connect` needs to be declared above and passed explicitly into the function. We use `local` rather than global variables.

examples/example-with-consul-connect/user-data-client.sh
7hacker: no that doesnt happen yet, but should happen along with tests to verify the intentions

examples/example-with-consul-connect/variables.tf
```diff
@@ -0,0 +1,61 @@
...
+variable "cluster_tag_key" {
+  description = "The tag the EC2 Instances will look for to automatically discover each other and form a cluster."
+  type        = string
+  default     = "consul-servers"
+}
+
+variable "ssh_key_name" {
+  description = "The name of an EC2 Key Pair that can be used to SSH to the EC2 Instances in this cluster. Set to an empty string to not associate a Key Pair."
+  type        = string
+  default     = nt-trial
```
7hacker, line 48: set this back to default = null

MatthiasScholz:
@7hacker - I would like to jump in and help getting the PR merged. Any support needed?

examples/example-with-consul-connect/variables.tf
MatthiasScholz: Takled in #202.

modules/run-consul/run-consul
MatthiasScholz: Takled in #202.

modules/run-consul/run-consul
MatthiasScholz: Takled in #202.

brikis98:
Closing this PR due to inactivity.

PR #175, opened 2020-05-19
Add consul snapshot flags (Consul enterprise)

Lucretius:
Allow option for enabling snapshot agent, as well as bucket specification (where the snapshots are placed) and bucket path prefix.  This will only occur if IAM setup is enabled.  If someone is using a custom IAM role, they can just set up the required policy on their custom role.

Permissions for the snapshots match what is specified in the [Consul snapshot agent docs](https://www.consul.io/docs/commands/snapshot/agent.html#s3-required-permissions).

### DIFF
--- a/modules/consul-iam-policies/main.tf
+++ b/modules/consul-iam-policies/main.tf
@@ -35,13 +35,14 @@ data "aws_iam_policy_document" "auto_discover_cluster" {
 
 
 resource "aws_iam_role_policy" "snapshot_agent_to_s3" {
-  count  = var.enabled ? 1 : 0
+  count  = "${var.enabled && var.enable_snapshot_agent ? 1 : 0}"
   name   = "consul-snapshot-agent"
   role   = var.iam_role_id
   policy = data.aws_iam_policy_document.snapshot_agent_to_s3.json
 }
 
 data "aws_iam_policy_document" "snapshot_agent_to_s3" {
+  count = "${var.enabled && var.enable_snapshot_agent ? 1 : 0}"
   statement {
     effect = "Allow"
 
--- a/modules/consul-iam-policies/main.tf
+++ b/modules/consul-iam-policies/main.tf
@@ -62,6 +62,6 @@ data "aws_iam_policy_document" "snapshot_agent_to_s3" {
       "s3:DeleteObject"
     ]
 
-    resources = ["arn:aws:s3:::${var.snapshot_agent_bucket}/${var.snapshot_agent_s3_key_prefix}*.snap"]
+    resources = ["arn:aws:s3:::${var.snapshot_agent_bucket}/${var.snapshot_agent_s3_key_prefix}/*.snap"]
   }
 }
--- a/modules/consul-iam-policies/main.tf
+++ b/modules/consul-iam-policies/main.tf
@@ -35,14 +35,14 @@ data "aws_iam_policy_document" "auto_discover_cluster" {
 
 
 resource "aws_iam_role_policy" "snapshot_agent_to_s3" {
-  count  = "${var.enabled && var.enable_snapshot_agent ? 1 : 0}"
+  count  = ${var.enabled && var.enable_snapshot_agent} ? 1 : 0
   name   = "consul-snapshot-agent"
   role   = var.iam_role_id
   policy = data.aws_iam_policy_document.snapshot_agent_to_s3[0].json
 }
 
 data "aws_iam_policy_document" "snapshot_agent_to_s3" {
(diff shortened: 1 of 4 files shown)


brikis98:
Thanks for the PR! I'm super booked this week, but will try to take a look in the next few days.

modules/consul-iam-policies/main.tf
```diff
@@ -31,3 +31,37 @@ data "aws_iam_policy_document" "auto_discover_cluster" {
   }
 }
 
+# Additionally, if desired, add the necessary policy to allow snapshots to s3 buckets
+
+
+resource "aws_iam_role_policy" "snapshot_agent_to_s3" {
+  count  = "${var.enabled && var.enable_snapshot_agent}" ? 1 : 0
```
brikis98, line 38:
```suggestion
  count  = var.enabled && var.enable_snapshot_agent ? 1 : 0
```

modules/consul-iam-policies/main.tf
```diff
@@ -31,3 +31,37 @@ data "aws_iam_policy_document" "auto_discover_cluster" {
...
+  name   = "consul-snapshot-agent"
+  role   = var.iam_role_id
+  policy = data.aws_iam_policy_document.snapshot_agent_to_s3[0].json
+}
+
+data "aws_iam_policy_document" "snapshot_agent_to_s3" {
+  count = "${var.enabled && var.enable_snapshot_agent}" ? 1 : 0
```
brikis98, line 45:
```suggestion
  count = var.enabled && var.enable_snapshot_agent ? 1 : 0
```

modules/consul-iam-policies/main.tf
```diff
@@ -31,3 +31,37 @@ data "aws_iam_policy_document" "auto_discover_cluster" {
...
+
+    resources = ["arn:aws:s3:::${var.snapshot_agent_bucket}"]
+  }
+
+  statement {
+    effect = "Allow"
+
+    actions = [
+      "s3:PutObject",
+      "s3:DeleteObject"
```
brikis98, line 62: Should this, plus the `List` operations above only be added if snapshot rotation is enabled?

Lucretius: Yes, these operations are only required when the snapshot agent is running.  This policy is pulled right from [this snapshot agent docs](https://www.consul.io/docs/commands/snapshot/agent.html#s3-required-permissions)

brikis98: Yes, very docs say:

`PutObject` is "Required for all operations" whereas `DeleteObject`, `ListBucket`, and `ListBucketVersions` are "Required only when snapshot rotation is enabled." And from what you're saying, `GetObject` is necessary to restore.

So the question is, does it make sense to grant all properties always, or should we add additional `enable_snapshot_rotation` and `enable_restore` flags that only add their corresponding permissions when enabled?

Lucretius: Ah I see what you mean, sorry.  The following is my opinion on the matter:

I see no problem leaving everything enabled as-is.  The reason being is that the user also needs to specify (or accept the default) the s3 bucket and s3 key prefix where the backups will be stored.  Unless, for some reason, the user is storing non-backup files in the same folder where they write the backups *and* has fear of those being somehow deleted (which I don't think is something anyone will be doing, nor will the snapshot agent do that) I think it is easier to just allow it all without needing an additional flag.  It would take a decent amount of self-sabotage for those additional permissions to cause issues.

brikis98: I think the reason to follow the "principal of least privilege" is not just about user error, but also malicious attackers. E.g., Removing the read permissions would ensure that an attacker who got access to the Consul server can't read the data that was backed up; removing delete permissions ensures they can't delete the backups.

Lucretius: I've changed the PR to only set PutObject as a permission.  That said I do not know how useful people will find this PR as my guess is the majority of them are using the snapshot rotation capabilities of the agent.

If that is desired, this PR is going to require some changing to support that.  I do not have the time to devote to this and I've also found a way around this, by adding the needed policies after creating the cluster using the `iam_role_id` output variable.  As a result, I do not need this PR anymore and am fine closing it.  If someone later would like to support the full snapshot agent capabilities within a PR, I think it would be best to save it for them.

brikis98: > by adding the needed policies after creating the cluster using the iam_role_id output variable. 

Yup, that's exactly why we export all those IDs/ARNs as output variables!

> As a result, I do not need this PR anymore and am fine closing it.

Roger!

Issue #103, opened 2018-11-28
Feature Request: Support Launch-Templates for t2/t3 unlimited configuration


eedwards-sk:
Launch Configurations do not support configuring t2.unlimited or t3 bursting

This is due to AWS only supporting configuration of these options through launch templates.

This is a feature request to add support to use launch templates rather than launch configurations.

This is known to be a backwards incompatible change, thus this may be something for a much later release (major).

eedwards-sk:
relevant issue: https://github.com/hashicorp/terraform-aws-vault/pull/97

looks like our best bet for getting this implemented will be with 0.12+

brikis98:
+1. Adding this with 0.12 should be a whole lot easier.

acsrujan:
Hey, can this be revisited? 

I'm not here for t3.unlimited but AWS raised an alarm saying ASG doesn't have a launch template.

robmorgan:
Hi @acsrujan we don't have any time to work on this at the moment, but we're happy to review any PRs if you want to look into it?

acsrujan:
Thanks @robmorgan 

Added a quick draft PR https://github.com/hashicorp/terraform-aws-consul/pull/176

Will mark it as ready for review after following guidelines and testing it whenever I get time.

PR #176, opened 2020-05-28
changes launch configuration to launch template

acsrujan:


acsrujan:
Hey, 

1. Yes. I left it in draft version for work in progress. I'll close it and raise one when it's ready with backward compatible version (if feasible).
Upgrade without downtime is not yet addressed. I'm considering if we create launch template, attach it to existing ASG. Still working on making it frictionless.
This might require a discussion here or https://github.com/hashicorp/terraform-aws-consul/issues/103

2. No, both from official documentation and personal experience.

Issue #177, opened 2020-07-07
Instance scale-in protection missing consul-cluster


mawa-jnd:
It is not possible to set the Instance scale-in protection for consul-cluster instances.
The used "aws_autoscaling_group" allows to set "protect_from_scale_in" but it is not provided by the module:
https://github.com/hashicorp/terraform-aws-consul/blob/master/modules/consul-cluster/main.tf#L14

This issue is more a kind of feature request though.

PR #178, opened 2020-07-07
Add instance scale-in protection for consul-cluster

mawa-jnd:
This change solve issue: https://github.com/hashicorp/terraform-aws-consul/issues/177

The changes are tested with an existing consul-cluster without scale-in protection. The optional scale-in protection is set to true.
The terraform plan output is here:
https://gist.github.com/mawa-jnd/476ab6680a68d6c4b1770a372cfff5aa

The change is:
- backward compatible
- has no downtime with default setup as changing the ASG will not automatically restart instances

### DIFF
--- a/modules/consul-cluster/main.tf
+++ b/modules/consul-cluster/main.tf
@@ -32,6 +32,8 @@ resource "aws_autoscaling_group" "autoscaling_group" {
 
   enabled_metrics = var.enabled_metrics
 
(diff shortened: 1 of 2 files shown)


brikis98:
Tests passed! Merging now.

brikis98:
https://github.com/hashicorp/terraform-aws-consul/releases/tag/v0.7.5

PR #179, opened 2020-07-10
Remove duplicate version block in SG Rules module

RetWolf:
After some brief searching in the GitHub issues I couldn't find any reasoning for this, so I assume its just a mistake. This change removes a duplicated Terraform block with both specifying the "required_version". 

I have not run tests for this change, if you would like me to please let me know and I can do so.

This change should not affect backwards compatibility or cause downtime.


brikis98:
Tests passed! Merging now.

Issue #181, opened 2020-07-31
Provider-aws latest release has a conflict with consul-cluster module


rafaelremondes:
In the last version of the  aws provider, `v3.0.0`, following conflicts appear for resource `resource/aws_autoscaling_group` when running a `terraform plan`

```
Error: "availability_zones": conflicts with vpc_zone_identifier

  on .terraform/modules/consul_cluster/terraform-aws-consul-0.7.6/modules/consul-cluster/main.tf line 14, in resource "aws_autoscaling_group" "autoscaling_group":
  14: resource "aws_autoscaling_group" "autoscaling_group" {



Error: "vpc_zone_identifier": conflicts with availability_zones

  on .terraform/modules/consul_cluster/terraform-aws-consul-0.7.6/modules/consul-cluster/main.tf line 14, in resource "aws_autoscaling_group" "autoscaling_group":
  14: resource "aws_autoscaling_group" "autoscaling_group" {
```

Related to this change:
https://github.com/terraform-providers/terraform-provider-aws/pull/12927

Only one of the two parameters(`vpc_zone_identifier`, `availability_zones`) should be passed right?

brikis98:
3.0.0 just came out, with some backwards incompatible changes, and we have not yet tested/upgraded this repo for it. As a short term workaround, you can pin your AWS provider to `2.x.x`. 

> Only one of the two parameters(`vpc_zone_identifier`, `availability_zones`) should be passed right?

I think we want to pass both, but we probably need to change the `default` for our corresponding input variables in `consul-cluster` to be `null`. Would you be up for a PR that makes that change?

rafaelremondes:
Hey @brikis98 

Yes, I pinned the `provider-aws` version and it worked.

Sure, I'll test it and open a PR. Thanks for the help

joshmello:
@rafaelremondes are you going to put in a pr? I am having the same issue.

andrew-womeldorf:
:raised_hands: Thank you @rafaelremondes @brikis98 !

brikis98:
Fixed by @rafaelremondes in https://github.com/hashicorp/terraform-aws-consul/releases/tag/v0.7.7.

PR #182, opened 2020-08-01
Edit default value for availability_zones and subnet_ids vars to null

rafaelremondes:
Related to [Issue 181](https://github.com/hashicorp/terraform-aws-consul/issues/181)

Newest `provider-aws` version only allows us to set one of the 2 variables: `availability_zones and subnet_ids` when creating AutoScalingGroups.

As per @brikis98 suggestion, I am changing the default value of these 2 variables to null so they're no longer conflicting, allowing to set either of them when using the `consul_cluster` module.


brikis98:
Tests passed! Merging now.

brikis98:
https://github.com/hashicorp/terraform-aws-consul/releases/tag/v0.7.7

PR #190, opened 2020-08-25
Add support for server hostname checks

yorinasub17:
This adds support for enabling `verify_server_hostname` to `true` in the consul server configuration. This is a recommended setting for production consul servers to ensure that agents can't promote themselves to become a server using the client certificate key (if `verify_server_hostname` is `false`, then any certificate that is validated by the RPC CA can become a server).

See https://www.consul.io/docs/agent/options#verify_server_hostname for more info.

### DIFF
--- a/modules/run-consul/run-consul
+++ b/modules/run-consul/run-consul
@@ -51,6 +51,7 @@ function print_usage {
   echo -e "  --ca-path\t\tPath to the directory of CA files used to verify outgoing connections. Optional. Must be specified with --enable-rpc-encryption."
   echo -e "  --cert-file-path\tPath to the certificate file used to verify incoming connections. Optional. Must be specified with --enable-rpc-encryption and --key-file-path."
   echo -e "  --key-file-path\tPath to the certificate key used to verify incoming connections. Optional. Must be specified with --enable-rpc-encryption and --cert-file-path."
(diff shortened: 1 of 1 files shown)


brikis98 (commented):
The code LGTM, but is this a common enough setting to add to `run-consul` for all users? Or should the users that need it [provide custom configuration as shown here](https://github.com/hashicorp/terraform-aws-consul/tree/master/modules/run-consul#overriding-the-configuration)?

yorinasub17:
>  is this a common enough setting to add to run-consul for all users?

According to [the docs](https://www.consul.io/docs/agent/encryption.html), this is recommended to set to true for any cluster that is configuring RPC encryption:

> If verify_server_hostname is set, then outgoing connections perform hostname verification. All servers must have a certificate valid for server.<datacenter>.<domain> or the client will reject the handshake. This is a new configuration as of 0.5.1, and it is used to prevent a compromised client from being able to restart in server mode and perform a MITM (Man-In-The-Middle) attack. **New deployments should set this to true**, and generate the proper certificates, but this is defaulted to false to avoid breaking existing deployments.

So I think this is something we should offer in the core. In fact, I originally wanted to set this to `true` without an opt-in flag, as a production cluster using RPC encryption should have this included. However, as the docs mention, it breaks existing clusters that don't have the proper certs, so I opted for a feature flag instead to maintain backwards compatibility.

brikis98 (approved):
Roger! Kicking off tests now.

brikis98:
Tests passed! Merging now.

brikis98:
https://github.com/hashicorp/terraform-aws-consul/releases/tag/v0.7.9

PR #191, opened 2020-09-28
Allow configuration of dnsmasq listen address separate from the consul IP address

ryandens:
This pull request allows the configuration of the dnsmasq `listen-address`. Previously, the only way to modify this also modified the consul IP address that dnsmasq routed to. This is relatively backward compatible. Users of the default consul-ip default value will see the same behavior, as the dnsmasq listen-address defaults to the same value. However, if a user was running consul on a different IP, and ran dnsmasq on that same IP to accommodate this script, they will need to update their usage of this script.

### DIFF
--- a/modules/install-dnsmasq/README.md
+++ b/modules/install-dnsmasq/README.md
@@ -50,6 +50,8 @@ The `install-dnsmasq` script accepts the following arguments:
 * `consul-ip IP`: The IP address to use for Consul. Optional. Default: `127.0.0.1`. This assumes a Consul agent is 
   running locally and connected to a Consul cluster.
 * `consul-dns-port PORT`: The port Consul uses for DNS requests. Optional. Default: `8600`.
+* `dnsmasq-listen-address IP`: The IP address for dnsmasq to listen on. Optional. Defaults to the value of `consul-ip`. 
+Make sure that the network interface you provide for the IP has already been configured before you pass it to dnsmasq.
 
 Example:
 
--- a/modules/install-dnsmasq/install-dnsmasq
+++ b/modules/install-dnsmasq/install-dnsmasq
@@ -10,6 +10,7 @@ set -e
 readonly DEFAULT_CONSUL_DOMAIN="consul"
 readonly DEFAULT_CONSUL_IP="127.0.0.1"
 readonly DEFAULT_CONSUL_DNS_PORT=8600
+readonly DEFAULT_DNSMASQ_LISTEN_ADDRESS=""
 
 readonly DNS_MASQ_CONFIG_DIR="/etc/dnsmasq.d"
 readonly CONSUL_DNS_MASQ_CONFIG_FILE="$DNS_MASQ_CONFIG_DIR/10-consul"
@@ -27,6 +28,7 @@ function print_usage {
   echo -e "  --consul-domain\tThe domain name to point to Consul. Optional. Default: $DEFAULT_CONSUL_DOMAIN."
   echo -e "  --consul-ip\t\tThe IP address to use for Consul. Optional. Default: $DEFAULT_CONSUL_IP."
   echo -e "  --consul-dns-port\tThe port Consul uses for DNS. Optional. Default: $DEFAULT_CONSUL_DNS_PORT."
+  echo -e "  --dnsmasq-listen-address\t\tThe IP address for dnsmasq to listen on. Optional. Defaults to the value of --consul-ip."
   echo
   echo "Example:"
   echo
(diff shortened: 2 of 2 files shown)


modules/install-dnsmasq/install-dnsmasq
```diff
@@ -98,15 +100,16 @@ function write_consul_config {
...
-  log_info "Configuring Dnsmasq to forward lookups of the '$consul_domain' domain to $consul_ip:$consul_port in $CONSUL_DNS_MASQ_CONFIG_FILE"
+  log_info "Configuring Dnsmasq to forward lookups of the '$consul_domain' domain on '$dnsmasq_listen_address' to $consul_ip:$consul_port in $CONSUL_DNS_MASQ_CONFIG_FILE"
   mkdir -p "$DNS_MASQ_CONFIG_DIR"
 
   sudo tee "$CONSUL_DNS_MASQ_CONFIG_FILE" <<EOF
 # Enable forward lookup of the '$consul_domain' domain:
 server=/${consul_domain}/${consul_ip}#${consul_port}
 
-listen-address=${consul_ip}
+listen-address=${dnsmasq_listen_address}
```
brikis98, line 112: Hm, it seems like we should keep this at the previous default of `consul_ip`.

That is, in the `install` function, we do the following:

1. By default, `dnsmasq_listen_address` is set to an empty string (`""`).
1. If a user passes in `--dnsmasq-listen-address`, we use their value for `dnsmasq_listen_address`. 
1. Otherwise, after the `while` loop to parse args, if `dnsmasq_listen_address` is still an empty string, then we set it to the `consul_ip` value.

This is fully backwards compatible and still allows `dnsmasq_listen_address` to be customized.

brikis98 (commented):
Thanks for the PR!

modules/install-dnsmasq/install-dnsmasq
ryandens: Thanks! This is a great way to ensure backward compatibility

modules/install-dnsmasq/install-dnsmasq
```diff
@@ -148,10 +157,14 @@ function install {
 
     shift
   done
+  
+  if [[ -z "$dnsmasq_listen_address" ]]; then
+    dnsmasq_listen_address = consul_ip
```
brikis98, line 162: I think this is invalid syntax. Should be:

```suggestion
    dnsmasq_listen_address="$consul_ip"
```

modules/install-dnsmasq/install-dnsmasq
```diff
@@ -27,6 +28,7 @@ function print_usage {
   echo -e "  --consul-domain\tThe domain name to point to Consul. Optional. Default: $DEFAULT_CONSUL_DOMAIN."
   echo -e "  --consul-ip\t\tThe IP address to use for Consul. Optional. Default: $DEFAULT_CONSUL_IP."
   echo -e "  --consul-dns-port\tThe port Consul uses for DNS. Optional. Default: $DEFAULT_CONSUL_DNS_PORT."
+  echo -e "  --dnsmasq-listen-address\t\tThe IP address for dnsmasq to listen on. Optional. Defaults to the value of --consul-ip"
```
brikis98, line 31: NIT:

```suggestion
  echo -e "  --dnsmasq-listen-address\t\tThe IP address for dnsmasq to listen on. Optional. Defaults to the value of --consul-ip."
```

modules/install-dnsmasq/README.md
```diff
@@ -50,6 +50,8 @@ The `install-dnsmasq` script accepts the following arguments:
 * `consul-ip IP`: The IP address to use for Consul. Optional. Default: `127.0.0.1`. This assumes a Consul agent is 
   running locally and connected to a Consul cluster.
 * `consul-dns-port PORT`: The port Consul uses for DNS requests. Optional. Default: `8600`.
+* `dnsmasq-listen-address IP`: The IP address for dnsmasq to listen on. Optional. Defaults to the value of `consul-ip`. 
+This assumes that the newtwork interface for the provided IP has been configured
```
brikis98, line 54: What does the last sentence mean?

ryandens: I left this to make sure the user knows that depending on the address they choose, they may need to set up the network interface at boot time prior to starting dnsmasq. An example use case is that if you're configuring dnsmasq to bind to a dummy interface. This could be more specific to my use-case than I realize, so I'm happy to remove it. I mostly didn't want users to think that they could simply put any IP address as a parameter of this script and that everything will work without extra configuration. Admittedly I'm no expert in networking on Linux machines, so I definitely could have some poorly chosen language here and am not opposed to doing more research and rephrasing if you think it's valuable

brikis98: Gotcha. Perhaps rephrasing to something like, "Make sure that the network interface you provide for the IP has already been configured before you pass it to dnsmasq" would be a bit clearer?

ryandens: Agreed!

brikis98 (approved):
LGTM, thanks! I'll kick off tests now.

brikis98:
Tests passed! Merging now.

brikis98:
https://github.com/hashicorp/terraform-aws-consul/releases/tag/v0.7.11

ryandens:
Thank you!

PR #192, opened 2020-10-15
Update repo to work with TF 0.13

brikis98:
1. Update `required_version` to `0.12.26` everywhere.
1. Add pre-commit hooks.
1. Update version numbers in CI build.
1. Run `go fmt`.
1. Fix folder copy in a Packer template. Our previous Circle CI Docker image used to check out the repo into a folder called `terraform-aws-consul`, so the Packer template could use a `file` provisioner to copy it from that folder name. Our new Circle CI Docker image uses a different folder name (e.g., `project`), so we have to update the Packer template to work even if it doesn't know the folder name.
1. Add retries for Packer builds for what looks like a transient build failure.

PR #193, opened 2020-10-15
Fix deploy step in build

brikis98:
In https://github.com/hashicorp/terraform-aws-consul/pull/192, I updated the build to use a different Docker image, with a newer version of Go that doesn't rely on everything running in the `/go/xxx` path. Unfortunately, I forgot to update the deploy portion of the build, which publishes AMIs, and was still relying on those paths. This PR fixes the deploy step. I created a temporary, internal release to validate it, as the deploy step doesn't run on branches, by default: https://github.com/hashicorp/terraform-aws-consul/releases/tag/v0.8.1-internal-test


brikis98:
Gah, WTH... Why does my PR get merged when I do an internal release of it???

PR #196, opened 2020-11-05
Add encrypted root device and additional EBS volumes

### LINKED ISSUE #160: Module does not support encrypted block ebs_block_device and root_block_device option
Can you add the option inside the `aws_launch_configuration` ressource on the module consul-cluster

https://github.com/hashicorp/terraform-aws-consul/blob/master/modules/consul-cluster/main.tf#L76
to encrypt the ebs_block_device and root_block_device 

Thanks

mirkop-mattr:
Fixes #160 

Add option to use an encrypted root volume

Allow for additional EBS volumes to be mounted

### DIFF
--- a/modules/consul-cluster/variables.tf
+++ b/modules/consul-cluster/variables.tf
@@ -166,11 +166,12 @@ variable "root_volume_delete_on_termination" {
 
 variable "root_volume_encrypted" {
   description = "Encrypt the volume at rest"
+  type        = bool
   default     = false
 }
 
 variable "ebs_block_device" {
-  description = "Additional EBS block devices to attach to autoscaling instances" 
+  description = "Additional EBS block devices to attach to autoscaling instances"
   type        = list(map(string))
   default     = []
 }
--- a/modules/consul-cluster/variables.tf
+++ b/modules/consul-cluster/variables.tf
@@ -164,6 +164,17 @@ variable "root_volume_delete_on_termination" {
   default     = true
 }
 
+variable "root_volume_encrypted" {
+  description = "Encrypt the volume at rest"
+  default     = false
+}
+
+variable "ebs_block_devices" {
+  description = "List of EBS volumes"
+  type        = list
+  default     = []
+}
+
 variable "wait_for_capacity_timeout" {
   description = "A maximum duration that Terraform should wait for ASG instances to be healthy before timing out. Setting this to '0' causes Terraform to skip all Capacity Waiting behavior."
   type        = string
--- a/modules/consul-cluster/variables.tf
+++ b/modules/consul-cluster/variables.tf
@@ -169,9 +169,9 @@ variable "root_volume_encrypted" {
   default     = false
 }
 
-variable "ebs_block_devices" {
-  description = "List of EBS volumes"
-  type        = list
+variable "ebs_block_device" {
+  description = "Additional EBS block devices to attach to autoscaling instances" 
+  type        = list(map(string))
   default     = []
 }
 
--- a/modules/consul-cluster/main.tf
+++ b/modules/consul-cluster/main.tf
@@ -109,8 +109,8 @@ resource "aws_launch_configuration" "launch_configuration" {
       volume_size           = lookup(ebs_block_device.value, "volume_size", null)
       volume_type           = lookup(ebs_block_device.value, "volume_type", null)
     }
-  }  
-  
+  }
+
   # Important note: whenever using a launch configuration with an auto scaling group, you must set
   # create_before_destroy = true. However, as soon as you set create_before_destroy = true in one resource, you must
   # also set it in every resource that it depends on, or you'll get an error about cyclic dependencies (especially when
--- a/modules/consul-cluster/main.tf
+++ b/modules/consul-cluster/main.tf
@@ -97,7 +97,7 @@ resource "aws_launch_configuration" "launch_configuration" {
     delete_on_termination = var.root_volume_delete_on_termination
     encrypted             = var.root_volume_encrypted
   }
-  
+
   dynamic "ebs_block_device" {
     for_each = var.ebs_block_devices
     content {
--- a/modules/consul-cluster/main.tf
+++ b/modules/consul-cluster/main.tf
@@ -102,7 +102,7 @@ resource "aws_launch_configuration" "launch_configuration" {
     for_each = var.ebs_block_devices
     content {
       delete_on_termination = lookup(ebs_block_device.value, "delete_on_termination", null)
-      device_name           = ebs_block_device.value.device_name
+      device_name           = lookup(ebs_block_device.value, "device_name", null)
       encrypted             = lookup(ebs_block_device.value, "encrypted", null)
       iops                  = lookup(ebs_block_device.value, "iops", null)
       snapshot_id           = lookup(ebs_block_device.value, "snapshot_id", null)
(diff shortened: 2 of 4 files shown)


mirkop-mattr:
Do you want me to create an example for an additional volume?

modules/consul-cluster/variables.tf
```diff
@@ -164,6 +164,17 @@ variable "root_volume_delete_on_termination" {
   default     = true
 }
 
+variable "root_volume_encrypted" {
+  description = "Encrypt the volume at rest"
+  default     = false
```
brikis98, line 169: Please set the `type` to `bool`

modules/consul-cluster/variables.tf
```diff
@@ -164,6 +164,17 @@ variable "root_volume_delete_on_termination" {
...
+}
+
+variable "ebs_block_devices" {
+  description = "List of EBS volumes"
+  type        = list
+  default     = []
```
brikis98, line 175: Please add a comment with an example of the value you can set this too.

modules/consul-cluster/variables.tf
```diff
@@ -164,6 +164,17 @@ variable "root_volume_delete_on_termination" {
...
 }
 
+variable "root_volume_encrypted" {
+  description = "Encrypt the volume at rest"
+  default     = false
+}
+
+variable "ebs_block_devices" {
+  description = "List of EBS volumes"
+  type        = list
```
brikis98, line 174: This should be a map, not a list so that (a) the `for_each` works and (b) each volume has a unique name/ID, so if a user removes one or adds one, it only affects that one, rather than all the items after it in a list.

modules/consul-cluster/variables.tf
```diff
@@ -164,6 +164,17 @@ variable "root_volume_delete_on_termination" {
   default     = true
 }
 
+variable "root_volume_encrypted" {
+  description = "Encrypt the volume at rest"
+  default     = false
+}
+
+variable "ebs_block_devices" {
+  description = "List of EBS volumes"
```
brikis98, line 173: Please update this description as follows:

1. Mention that this is the list of EBS volumes to create... But the user will have to attach and mount these volumes themselves: e.g., in a User Data script.
1. Specify that the type is a map from some unique name / ID for the volume (see comment below), which is used solely in this Terraform code, and that the values are objects with the fields supported by the `ebs_block_device` block in `aws_launch_configuration`. You may want to link to the `aws_launch_configuration` docs too.

brikis98 (commented):
Thanks for the PR!

Sanity check: is this type of EBS volume useful? EBS volumes you specify in a launch configuration are not "persistent" volumes that survive between redeploys of the EC2 instances in the ASG. So if you just need ephemeral disk storage that goes away when the instances are replaced, then this works fine, but if you are looking for persistent storage that will survive redeploys (e.g., the EBS volume is detached from an old instance and reattached to its replacement), this will not work.

If the `ebs_block_devices` param does work, then yes, it's worth updating one of the existing examples to use it. That way, the tests will execute that code. As it is now, I don't think this code works (see the comment on using a map instead of list so the `for_each` works).

mirkop-mattr:
> Thanks for the PR!
> 
> Sanity check: is this type of EBS volume useful? EBS volumes you specify in a launch configuration are not "persistent" volumes that survive between redeploys of the EC2 instances in the ASG. So if you just need ephemeral disk storage that goes away when the instances are replaced, then this works fine, but if you are looking for persistent storage that will survive redeploys (e.g., the EBS volume is detached from an old instance and reattached to its replacement), this will not work.
> 
> If the `ebs_block_devices` param does work, then yes, it's worth updating one of the existing examples to use it. That way, the tests will execute that code. As it is now, I don't think this code works (see the comment on using a map instead of list so the `for_each` works).

Hi @brikis98, thanks for your comments. Yes initially i thought those EBS volumes would persist and I was obviously wrong. Do you still think there is value of attaching additional EBS volumes rather than increasing the root volume size? I can think of using cheap storage for OS, high performance storage for consul data. Or avoiding issues of the OS filling up.
I have not found a simple way of having persistent disk attached ( similar to a Statefulset in Kubernetes) to store the consul data. It's a bit scary to think that if there is not a single EC2 instances running all your consul data is lost.

brikis98:
> > Thanks for the PR!
> > Sanity check: is this type of EBS volume useful? EBS volumes you specify in a launch configuration are not "persistent" volumes that survive between redeploys of the EC2 instances in the ASG. So if you just need ephemeral disk storage that goes away when the instances are replaced, then this works fine, but if you are looking for persistent storage that will survive redeploys (e.g., the EBS volume is detached from an old instance and reattached to its replacement), this will not work.
> > If the `ebs_block_devices` param does work, then yes, it's worth updating one of the existing examples to use it. That way, the tests will execute that code. As it is now, I don't think this code works (see the comment on using a map instead of list so the `for_each` works).
> 
> Hi @brikis98, thanks for your comments. Yes initially i thought those EBS volumes would persist and I was obviously wrong. Do you still think there is value of attaching additional EBS volumes rather than increasing the root volume size? I can think of using cheap storage for OS, high performance storage for consul data. Or avoiding issues of the OS filling up.
> I have not found a simple way of having persistent disk attached ( similar to a Statefulset in Kubernetes) to store the consul data. It's a bit scary to think that if there is not a single EC2 instances running all your consul data is lost.

Separate EBS volumes, even ephemeral ones, could be useful for a variety of reasons. That said, I worry many will assume they are persistent volumes, and that'll lead to data loss & confusion...

The way to add persistent volumes is to create them outside the ASG via `aws_ebs_volume` resources. You then add code to your EC2 instances to attach volumes when they are booting. But there's more complexity there, as you can only attach an EBS volume in the same AZ as the EC2 instance, whereas an ASG could launch an instance in any AZ...

perlboy:
Just want to +1 on this PR. While the volumes may be ephemeral, allowing for encryption to be enabled on them allows for global policy within organisations (ie. "Disk Encryption must be enabled") to be maintained. I accept that in the context of this specific module that may not be seen as relevant but adding this support inline helps technical folk skate past the somewhat less aware ("What's ephemeral?") compliance folk and "earn" their tick on that policy checkpoint.

Also @robmorgan [this addition snuck](https://github.com/hashicorp/terraform-aws-consul/pull/196/files#diff-6c605a7da1f7d7fe7888c86e0165164c3a93d4059ac17be4e64c6cfbe6dc341bR85) in still. I support that change too but it's not aligned with the PR.

mirkop-mattr:
I have actually refactored the code completely, but not yet updated the PR yet. I am not happy with using an autoscaling group for consul and changed it to use a fixed number of instances with persistent storage.

perlboy:
@mirkop-mattr So are you intending to abandon this PR then? To be honest my interest here is purely centered around the root volume within the ASG which is a fairly trivial change versus the addition of dynamic EBS volumes.

mirkop-mattr:
I was not quite sure what to do as I was still experimenting with changing it to fixed instances. If it helps you I am happy to break this up and change the PR to enabling encryption for the root volume only.

perlboy:
> I was not quite sure what to do as I was still experimenting with changing it to fixed instances. If it helps you I am happy to break this up and change the PR to enabling encryption for the root volume only.

 :+1: (selfishly) Yes, it would be helpful otherwise I would look at forking and creating my own PR with just the root volume part added.

mirkop-mattr:
Leave that with me. I will create two PRs: one for root volume encryption and another one for ubuntu 20.04 support.

PR #197, opened 2020-11-09
Update systemd-resolved script for Ubuntu 20.04

mirkop-mattr:
Systemd is also available on Ubuntu 20.04. The updates will allow the provisioning script to run.

### DIFF
--- a/test/consul_cluster_test.go
+++ b/test/consul_cluster_test.go
@@ -17,6 +17,11 @@ func TestConsulClusterWithUbuntu18Ami(t *testing.T) {
 }
 
 // Test the example in the root folder
+func TestConsulClusterWithUbuntu20Ami(t *testing.T) {
+	t.Parallel()
+	runConsulClusterTest(t, "ubuntu20-ami", ".", "../examples/consul-ami/consul.json", "ubuntu", "")
+}
+// Test the example in the root folder
 func TestConsulClusterWithAmazonLinuxAmi(t *testing.T) {
 	t.Parallel()
 	runConsulClusterTest(t, "amazon-linux-2-ami", ".", "../examples/consul-ami/consul.json", "ec2-user", "")
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ To deploy Consul clients for production using this repo:
 1. Use the [install-consul module](https://github.com/hashicorp/terraform-aws-consul/tree/master/modules/install-consul) to install Consul alongside your application code.
 1. Before booting your app, execute the [run-consul script](https://github.com/hashicorp/terraform-aws-consul/tree/master/modules/run-consul) with `--client` flag.
 1. Your app can now use the local Consul agent for service discovery and key/value storage.
-1. Optionally, you can use the [install-dnsmasq module](https://github.com/hashicorp/terraform-aws-consul/tree/master/modules/install-dnsmasq) for Ubuntu 16.04 and Amazon Linux 2 or [setup-systemd-resolved](https://github.com/hashicorp/terraform-aws-consul/tree/master/modules/setup-systemd-resolved) for Ubuntu 18.04 to configure Consul as the DNS for a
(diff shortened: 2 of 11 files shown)


robmorgan:
Thanks @mirkop-mattr! How did you test this? I guess we should explitity add Ubuntu 20.04 as a target to [our tests](https://github.com/hashicorp/terraform-aws-consul/blob/master/test/consul_cluster_test.go#L8) and Packer templates.

mirkop-mattr:
> Thanks @mirkop-mattr! How did you test this? I guess we should explitity add Ubuntu 20.04 as a target to [our tests](https://github.com/hashicorp/terraform-aws-consul/blob/master/test/consul_cluster_test.go#L8) and Packer templates.

Apologies I can try to run the tests tomorrow. I have actually build a Ubuntu 20.04 AMI myself, but with a packer template that did not include 16.04 anymore and looked slightly different. So yes it is probably a good idea to run the testsuite.

robmorgan:
@mirkop-mattr okay as part of this PR I think we also need to include examples + tests for Ubuntu 20.04 otherwise there's no way for our automated test suite to continually validate that this change works correctly. Are you able to add Ubuntu 20.04 to the Packer templates and tests?

Alternatively, you could also try another approach here: https://github.com/hashicorp/terraform-aws-consul/pull/197/files#diff-6c605a7da1f7d7fe7888c86e0165164c3a93d4059ac17be4e64c6cfbe6dc341bR85 that somehow works on all Ubuntu versions greater than 18.04. That way we won't have to explicitly list each distribution. Note: I'm assuming that's probably a bit more difficult than what I've proposed above.

brikis98 (commented):
Thanks for the PR! 

Before merging, we'd need the following:

1. Update the example Packer template to add an Ubuntu 20.04 builder.
1. Update the automated tests to add a test that uses that builder.
1. Update all the docs to say we're now testing with Ubuntu 20.04. We mention 16.04 and 18.04 in a bunch of places in the READMEs and comments, so we need to add 20.04 to that list.
1. I would then recommend building a new AMI and trying to run it manually to see what happens. The automated tests will of course do this too, but if there's some change in 20.04 that breaks the behavior, it'll be easier to tell by running it manually and looking at the logs, than to have the tests fail because Consul never booted correctly.

brikis98:
Doh, race condition! Looks like I duplicated @robmorgan's comments. Sorry about that 😁

mirkop-mattr:
Thanks for the feedback! Do you want me to update the circle ci configuration as well?

mirkop-mattr:
> @mirkop-mattr okay as part of this PR I think we also need to include examples + tests for Ubuntu 20.04 otherwise there's no way for our automated test suite to continually validate that this change works correctly. Are you able to add Ubuntu 20.04 to the Packer templates and tests?
> 
> Alternatively, you could also try another approach here: https://github.com/hashicorp/terraform-aws-consul/pull/197/files#diff-6c605a7da1f7d7fe7888c86e0165164c3a93d4059ac17be4e64c6cfbe6dc341bR85 that somehow works on all Ubuntu versions greater than 18.04. That way we won't have to explicitly list each distribution. Note: I'm assuming that's probably a bit more difficult than what I've proposed above.

Do you have any suggestions for another approach? I could check if the system we are running on is using systemd.

robcxyz:
Any updates on merging this?

test/consul_cluster_test.go
```diff
@@ -16,6 +16,11 @@ func TestConsulClusterWithUbuntu18Ami(t *testing.T) {
 	runConsulClusterTest(t, "ubuntu18-ami", ".", "../examples/consul-ami/consul.json", "ubuntu", "")
 }
 
+// Test the example in the root folder
+func TestConsulClusterWithUbuntu20Ami(t *testing.T) {
+	t.Parallel()
+	runConsulClusterTest(t, "ubuntu20-ami", ".", "../examples/consul-ami/consul.json", "ubuntu", "")
+}
```
brikis98, line 23: NIT: newline missing between these functions.

brikis98 (commented):
Sorry for the delay. We're super buried. Trying to chip away at my PR backlog!

Overall, the changes LGTM. Could you pull in the latest from `master`, and then I can kick off tests?

> Do you want me to update the circle ci configuration as well?

Update it in what way?

1FastSTi:
@mirkop-mattr , can you sign that CLA?

mirkop-mattr:
Any chance we can kick off the signing of the CLA again? I have changed email addresses for this account. Or shall I raise a new PR?

brikis98:
I don't see any way to do it other than a new PR. Thx!

mirkop-mattr:
Closing this and raising https://github.com/hashicorp/terraform-aws-consul/pull/217 using different github account.

Issue #209, opened 2021-01-26
Intermittent test failure


brikis98:
Some of the tests are intermittently failing with this error:

```
TestConsulClusterWithCustomASGRoleAmazonLinuxAmi 2021-01-26T12:15:50Z command.go:168: [1m[31mError: [0m[0m[1mError authorizing security group rule type ingress: InvalidGroup.NotFound: The security group 'sg-0e5cb357476add9f8' does not exist
```

Probably some sort of Terraform eventual consistency bug.

PR #208, opened 2021-01-26
Update agent configuration

anouarchattouna:
* bump `min_packer_version` -> `1.5.4`
* bump `consul_version` -> `1.9.2`
* update consul service configuration
  * Replace deprecated `ui` field by `ui_config` object: https://www.consul.io/docs/agent/options\#ui-1
  * Add `telemetry { disable_compat_1.9 = true }` to disable metrics deprecated in `v1.9`: https://www.consul.io/docs/agent/options\#telemetry-disable_compat_1.9

PR #212, opened 2021-02-25
Add more retries for entreprise test case

anouarchattouna:
For more details see #208 and #209

### DIFF
--- a/test/consul_helpers.go
+++ b/test/consul_helpers.go
@@ -53,10 +53,10 @@ func runConsulClusterTest(t *testing.T, packerBuildName string, examplesFolder s
 
 func runConsulClusterTestWithVars(t *testing.T, packerBuildName string, examplesFolder string, packerTemplatePath string, sshUser string, terraformVarsMerge map[string]interface{}, enterpriseUrl string) {
 	// Uncomment any of the following to skip that section during the test
-	//os.Setenv("SKIP_setup_ami", "true")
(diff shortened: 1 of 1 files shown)


test/consul_helpers.go
```diff
@@ -196,8 +196,7 @@ func checkEnterpriseInstall(t *testing.T, asgNameOutputVar string, terratestOpti
 		SshUserName: sshUser,
 		SshKeyPair:  keyPair.KeyPair,
 	}
-
-	maxRetries := 10
+	maxRetries := 20
```
ina-stoyanova, line 199: just curious - is there any derivation behind this value? I must be honest though - I'm lacking context, so excuse me if the question had been discussed before!

ina-stoyanova (approved):

test/consul_helpers.go
anouarchattouna: Actually, the idea is to add more retries for the `Check Enterprise Install` which is failing on `circleci` like you can see in this [comment](https://github.com/hashicorp/terraform-aws-consul/pull/208#issuecomment-784114002).
BTW: we are trying to find a way to fix #209

brikis98 (approved):
Let's give it a try! I'll kick off tests shortly.

brikis98:
Tests passed. Re-running to see if they are more stable now, as this was an intermittent issue.

brikis98:
Passed again! Merging now, thanks!

Issue #210, opened 2021-01-29
Could not find consul command. Aborting.


krishanhettihewa:
Today I just happened to noticed, though **consul-install** is successful ran, checking the version of the consul might fail, if we call this module remotely, such as by using Packer. 

The reason is, when command executed it just executed from the currently login user (E.g ubuntu). There is a possibility that normal user like `ubuntu` (None sudo user), might not have permissions to execute **/opt/consul/bin/consul**. Even Though we have the softlink which points `/usr/local/bin/consul >  /opt/consul/bin/consul`, due to lack of permissions to the `/opt/consul/bin/consul` the command get failed and because of this EXIT status get 1. 

This is might happen most probably, if we use hardened base image. I have noticed it works well when using general ubuntu base image, but not working when we uses our hardened base image.

 if command -v consul; then
    log_info "Consul install complete!";
  else
    log_info "Could not find consul command. Aborting.";
    exit 1;
  fi

https://github.com/hashicorp/terraform-aws-consul/blob/ae88203ccb739ad67448812025b6788a5737f607/modules/install-consul/install-consul#L303

brikis98:
Ah, interesting. So do we need a `sudo` in that command?

krishanhettihewa:
Thanks for the response @brikis98. Yes I hope so. it works well, with sudo permissions.

brikis98:
OK, would you be up for a quick PR to make that change?

yardbirdsax:
@brikis98 @krishanhettihewa I fixed this in the PR linked above, albeit in a slightly different way, which was to grant read / execute access to everyone on the Consul binary install path. This way the script works as intended (I tested it while calling it with Packer) and also if someone logs in to an instance as a different user they can still execute the Consul binary for troubleshooting / admin purposes.

yardbirdsax:
Actually please hold, there might be another issue.

yardbirdsax:
Okay we're good now. I ended up going the `sudo` route after all. I was having to tweak all sorts of permissions to get it working and in the end I didn't want to assume that those permissions should be there on someone's hardened image. This way the script works and it's left up to the user to determine the correct permissions for their needs.

PR #213, opened 2021-03-26
Fix for #210: can't find Consul binary when executing through Packer

yardbirdsax:
The install-consul script now grants read and execute access to the directory where the Consul binary is installed. This fixes the issue reported under #210, while also allowing users who sign in under a different login to use the Consul binary for administrative purposes if necessary.

### DIFF
--- a/modules/install-consul/install-consul
+++ b/modules/install-consul/install-consul
@@ -300,7 +300,7 @@ function install {
     install_tls_certificates "$path" "$user" "$ca_file_path" "$cert_file_path" "$key_file_path"
   fi
 
-  if command -v consul; then
+  if sudo bash -c "command -v consul"; then
     log_info "Consul install complete!";
   else
     log_info "Could not find consul command. Aborting.";
--- a/modules/install-consul/install-consul
+++ b/modules/install-consul/install-consul
@@ -302,7 +302,7 @@ function install {
 
   # This command is run with sudo rights so that it will succeed in cases where image hardening alters the permissions
   # on the directories where Consul may be installed. See https://github.com/hashicorp/terraform-aws-consul/issues/210.
-  if sudo bash -c "command -v consul"; then
+  if sudo -E PATH=$PATH bash -c "command -v consul"; then
     log_info "Consul install complete!";
   else
     log_info "Could not find consul command. Aborting.";
--- a/modules/install-consul/install-consul
+++ b/modules/install-consul/install-consul
@@ -300,6 +300,8 @@ function install {
     install_tls_certificates "$path" "$user" "$ca_file_path" "$cert_file_path" "$key_file_path"
   fi
 
+  # This command is run with sudo rights so that it will succeed in cases where image hardening alters the permissions
+  # on the directories where Consul may be installed. See https://github.com/hashicorp/terraform-aws-consul/issues/210.
   if sudo bash -c "command -v consul"; then
     log_info "Consul install complete!";
   else


yardbirdsax:
As per notes on #210 I ended up taking a different approach so as not to arbitrarily mess with permissions on the image's filesystem. I can imagine cases where a hardened image has very specific ACLs applied and didn't want to assume they should be undone.

modules/install-consul/install-consul
```diff
@@ -300,7 +300,7 @@ function install {
     install_tls_certificates "$path" "$user" "$ca_file_path" "$cert_file_path" "$key_file_path"
   fi
 
-  if command -v consul; then
+  if sudo bash -c "command -v consul"; then
```
brikis98, line 303: NIT: could you add a comment above this line explaining why we need `sudo` and linking back to #210?

brikis98 (commented):
Thank you for the PR! 👍 

Just one minor tweak and I can kick off tests.

modules/install-consul/install-consul
yardbirdsax: Done!

brikis98 (approved):
Nice, thank you! I'll kick off tests now.

brikis98:
Hm, something about this update is causing the Packer builds to fail at test time, at least on Amazon Linux:

```
TestConsulClusterWithAmazonLinuxAmi 2021-03-30T10:08:41Z command.go:168: 1617098921,,ui,error,==> amazon-linux-2-ami: 2021-03-30 10:08:41 [INFO] [install-consul] Could not find consul command. Aborting.
TestConsulClusterWithAmazonLinuxAmi 2021-03-30T10:08:41Z command.go:168: 1617098921,,ui,say,==> amazon-linux-2-ami: Provisioning step had errors: Running the cleanup provisioner%!(PACKER_COMMA) if present...
TestConsulClusterWithAmazonLinuxAmi 2021-03-30T10:08:41Z command.go:168: 1617098921,,ui,say,==> amazon-linux-2-ami: Terminating the source AWS instance...
```

My guess is that the path we install it into is not in `PATH` on Amazon Linux for the `sudo` user?

yardbirdsax:
Hmmm... that's odd indeed. Is there an easy way I can run the same Packer
build command locally that you do in your pipeline? I spent the last 10
minutes trying to unwind that and while I'm sure I could eventually I'm
hoping there's an easier way. :)

brikis98:
It's not super obvious, but you can see the command in the logs:

```
TestConsulClusterWithAmazonLinuxAmi 2021-03-30T10:06:40Z packer.go:91: Running Packer to generate a custom artifact for template ../examples/consul-ami/consul.json
TestConsulClusterWithAmazonLinuxAmi 2021-03-30T10:06:40Z retry.go:72: packer [build -machine-readable -var aws_region=ap-south-1 -only=amazon-linux-2-ami ../examples/consul-ami/consul.json]
```

So that becomes:

```bash
packer build \
  -machine-readable \
  -var aws_region=ap-south-1 \
  -only=amazon-linux-2-ami \
  ../examples/consul-ami/consul.json
```

yardbirdsax:
Thanks, I've started looking at this. The behavior is really strange since
`/usr/local/bin` is in the PATH of the sudo user.

yardbirdsax:
Turns out this did seem to be some kind of PATH tomfoolery. Manually passing in the user's current PATH value seems to have fixed it and was the simplest way I could determine to do so. I debated putting logic in the script to only do this under AL, but in the end it didn't appear to break other distros so I left it generic.

brikis98 (commented):
How odd! Thx for the update. I'll re-run the tests once more.

brikis98:
Hit a weird transient issue on the first test run, but it passed on re-run. The issue is unrelated to this PR, so I'm good to merge. Thanks!

brikis98:
https://github.com/hashicorp/terraform-aws-consul/releases/tag/v0.8.6

PR #221, opened 2021-04-28
Enhancement to support ACL bootstrapping

### LINKED ISSUE #119: Add ACL system config on `run_consul` on consul >= 1.4.0
I miss some options to configure and bootstrap ACL system since Consul v1.4.0 and later (https://www.consul.io/docs/agent/acl-system.html).

yardbirdsax:
Fixes #119 

## Functional Changes

- The `consul-iam-policies` module now includes permissions for reading / writing SSM parameters when a new variable named `acl_store_type` is set to "ssm".
- The `consul-cluster` module now includes an input variable named `acl_store_type`, which is used to indicate where created ACL tokens should be stored. Currently this only supports the value "ssm", but could support other methods in the future. This value is only used to pass on to the `consul-iam-policies` module.
- The example Packer file now includes a step to install the [`bash-commons`](https://github.com/gruntwork-io/bash-commons) library, which is used in cases where ACLs are enabled.
- Several new common bash functions were created to enable creation of ACL tokens and reading / writing of SSM parameters in support of enabling ACLs in created Consul clusters.  
- The `run-consul` script now accepts a switch parameter `--enable-acl`, which if set will cause the script to generate a root ACL token upon cluster provision, and store it as a specially named SSM parameter. The bootstrap command is executed on only one of the server instances; other server instances and client instances will wait for the SSM parameter to become available, then use it to create their own agent tokens with minimal rights. (Note: while this currently only supports SSM, it was written internally in a way that future expansion to other stores would be possible.)

## Testing / Refactoring related changes 

- Multiple Terratest related functions have been refactored to support the testing of ACL enabled clusters.
- Several bash functions have been moved to the new `consul-commons.sh` file so as to make testing them independent of running the `run-consul` script easier.

## Known Issues

- Some of the tests are failing, specifically the encryption enabled ones and the enterprise ones. I'm looking at these to get them corrected but wanted to get this up for review.

Test results: https://gist.github.com/yardbirdsax/e08efec4799c7e286ce093b5dfee89c3#file-terratest-logs-consul-acl-log

### DIFF
--- a/modules/run-consul/run-consul
+++ b/modules/run-consul/run-consul
@@ -53,6 +53,7 @@ function print_usage {
   echo -e "  --environment\t\tA single environment variable in the key/value pair form 'KEY=\"val\"' to pass to Consul as environment variable when starting it up. Repeat this option for additional variables. Optional."
   echo -e "  --skip-consul-config\tIf this flag is set, don't generate a Consul configuration file. Optional. Default is false."
   echo -e "  --recursor\tThis flag provides address of upstream DNS server that is used to recursively resolve queries if they are not inside the service domain for Consul. Repeat this option for additional variables. Optional."
+  echo -e "  --enable-acl\tThis flag will cause the cluster to bootstrap a root ACL token upon first provision, which is then stored in AWS SSM for other cluster members to retrieve."
   echo
   echo "Options for Consul Autopilot:"
   echo
@@ -312,6 +313,8 @@ EOF
 )
   fi
 
+  # TODO: Add step to add ACL section if --enable-acl is set, including token section if client == true
+
   log_info "Creating default Consul configuration"
   local default_config_json
   default_config_json=$(cat <<EOF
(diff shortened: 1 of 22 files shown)


modules/run-consul/run-consul
```diff
@@ -656,11 +318,59 @@ function run {
...
+      bootstrap_token=$(read_acl_token $cluster_tag_value "bootstrap" $aws_region "ssm")
+    fi
+
+    # Generate agent policy dynamically and write it
+    log_info "Creating agent policy and token"
+    local -r agent_policy=$(generate_node_acl_policy $local_hostname)
+    write_acl_policy "$instance_id" "$agent_policy" "$bootstrap_token"
+
+    # Generate agent token and persist it
+    local -r agent_token=$(generate_token "$instance_id" "$instance_id agent policy" $bootstrap_token)
```
mr-miles, line 370: Did you know you can generate a token with -node-identity "$instance_id" and it will get the policy you want without creating it explicitly? (was looking because i just did something similar to this and yours is way better!)

yardbirdsax: I did not know about that! It looks like it was added in Consul 1.8.1, so I wonder if it would be better to manually generate it since some folks might still be on an older version?

mr-miles: hmm good point, hadnt realised it was as new as that.

any chance of switching on version/option-flag?  I found the node identity option useful to avoid having tons of mechanical acl policies to wade through: when every agent has its own explicit policy there ends up being tons

modules/run-consul/run-consul
```diff
@@ -656,11 +318,59 @@ function run {
...
       "${recursors[@]}"
   fi
 
+  log_info "Creating Consul systemd file"
   generate_systemd_config "$SYSTEMD_CONFIG_PATH" "$config_dir" "$data_dir" "$systemd_stdout" "$systemd_stderr" "$bin_dir" "$user" "${environment[@]}"
+  log_info "Starting Consul"
   start_consul
+
+  if [[ "$enable_acl" == "true" ]]; then
+    local -r asg_name=$(aws_wrapper_get_asg_name $MAX_RETRIES $SLEEP_BETWEEN_RETRIES_SEC)
```
mr-miles, line 331: sorry one other question - and i have low knowledge of bash so i don't know if this is plausible.

i have an equivalent set of code (up to 364), basically using vault rather than ssm for the token storage.

i can see that you pass the "ssm" parameter so other stores could be added in the future.  but each of those functions is then going to have a switch statement in it.  i was wondering whether it is straightforward to just load the functions used based on a parameter.  Something like:

if [[ "$enable_acl" == "true" ]]; then
   if [[ "$acl_provider_path" != "" ]]; then
     source $acl_provider_path;
   else
     source ./aws-ssm-bootstrapper
   fi
fi

providing a name rather than a path would also work - other than ssm and vault, i can't think of other common things one might want to do, but i guess you can't rule out some kind of nfs share or other non-aws-native route

modules/run-consul/run-consul
yardbirdsax: We could make it so that if the Consul version is higher than 1.8.1 it will use that option, though I'd have to rethink the token generation function since that's specific to generating agent tokens (as opposed to the function now which is generic for any token given a policy). The bloat of agent policies is definitely a thing to consider, especially since ASG instances are not permanent things. Let me see how hard this would be, but I think we can include it here pretty easily.

modules/run-consul/run-consul
yardbirdsax: Hmm... so the idea your suggesting is that the `run-consul` script have a path for the acl storage provider, source that file dynamically, then call some known function name (like the current `write_acl_token` one)? It's certainly doable, I just get a little antsy about dynamically sourcing some random file, since with bash you can't really do things like enforce interfaces. The current idea is that the `write_acl_token` function is the entrypoint for writing tokens to durable storage, and that it would be enhanced over time to support multiple ones (like Vault, S3, etc) as needed.

yardbirdsax: Oh and should also note that, once another provider is required, the `run-consul` script would also be enhanced to have a parameter named something like `--acl-storage-type`, which would then be passed in to the `write_acl_token` function (vs right now where it's hard coded to pass 'ssm').

mr-miles: I've ping'd you a PR and comment, so you had a diff to look at rather than my unclear explanations

modules/run-consul/run-consul
yardbirdsax: I've made this update though I wasn't able to test in a lower version because of some configuration challenges. Now it will use the new method when the version is 1.8.1 or higher. Thanks for the suggestion!

modules/run-consul/run-consul
yardbirdsax: Thanks for the PR, that definitely helped me understand. Let me see what I can do there as I do agree it's a cleaner approach.

yardbirdsax:
Let me know if there's anything else required here? Thanks!

yardbirdsax:
Howdy, just wanted to ping here since I still have this PR outstanding. Was there anything else I need to do to get it closed out?

PR #222, opened 2021-05-05
Update install-consul logic to determine CPU arch which makes it run on Graviton based EC2 instances

izzyleung:
# Changes
- Include logic to determine binary architecture from returned value from `uname -m`

# Verification
- Tested on a `c6g.*` EC2 instance running Amazon Linux 2
```sh
$ uname -m
aarch64
```
- Tested on a Raspberry Pi 3B running Raspbian
```sh
$ uname -m
armv7l
```
- Tested on some good 'ol `amd64` Linux
```sh
$ uname -m
x86_64
```

### DIFF
--- a/modules/install-consul/install-consul
+++ b/modules/install-consul/install-consul
@@ -166,7 +166,8 @@ function fetch_binary {
   local -r version="$1"
   local download_url="$2"
 
-  local -r cpu_arch="$(uname -m)"
+  local cpu_arch
+  cpu_arch="$(uname -m)"
   local binary_arch=""
   case "$cpu_arch" in
     x86_64)
@@ -191,7 +192,8 @@ function fetch_binary {
       # $ uname -m
       # $ readelf -a /proc/self/exe | grep -q -c Tag_ABI_VFP_args && echo "armhf" || echo "armel"
       #
-      local -r vfp_tag="$(readelf -a /proc/self/exe | grep -q -c Tag_ABI_VFP_args)"
+      local vfp_tag
+      vfp_tag="$(readelf -a /proc/self/exe | grep -q -c Tag_ABI_VFP_args)"
       if [[ ! $vfp_tag  ]]; then
         binary_arch="armelv5"
       else
--- a/modules/install-consul/install-consul
+++ b/modules/install-consul/install-consul
@@ -166,8 +166,46 @@ function fetch_binary {
   local -r version="$1"
   local download_url="$2"
 
-  if [[ -z "$download_url" && -n "$version" ]];  then
-    download_url="https://releases.hashicorp.com/consul/${version}/consul_${version}_linux_amd64.zip"
+  local -r cpu_arch="$(uname -m)"
+  local binary_arch=""
+  case "$cpu_arch" in
+    x86_64)
+      binary_arch="amd64"
+      ;;
+    x86)
+      binary_arch="386"
+      ;;
+    aarch64)
+      binary_arch="arm64"
+      ;;
+    arm*)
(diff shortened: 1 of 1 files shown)


modules/install-consul/install-consul
```diff
@@ -166,8 +166,46 @@ function fetch_binary {
...
+      #
+      # Use Armelv5 for all 32-bit armel systems
+      # Use Armhfv6 for all armhf systems with v6+ architecture
+      # Use Arm64 for all v8 64-bit architectures
+      # The following commands can help determine the right version for your system:
+      #
+      # $ uname -m
+      # $ readelf -a /proc/self/exe | grep -q -c Tag_ABI_VFP_args && echo "armhf" || echo "armel"
+      #
+      local -r vfp_tag="$(readelf -a /proc/self/exe | grep -q -c Tag_ABI_VFP_args)"
```
izzyleung, line 194: Further readings: https://blogs.oracle.com/jtc/is-it-armhf-or-armel

modules/install-consul/install-consul
```diff
@@ -166,8 +166,46 @@ function fetch_binary {
   local -r version="$1"
   local download_url="$2"
 
-  if [[ -z "$download_url" && -n "$version" ]];  then
-    download_url="https://releases.hashicorp.com/consul/${version}/consul_${version}_linux_amd64.zip"
+  local -r cpu_arch="$(uname -m)"
```
brikis98, line 169: Declare variables and use subshells in separate statements, as otherwise, the exit code from the subshell will be lost, and we only get the exit code from `local`.

```suggestion
  local cpu_arch
  cpu_arch="$(uname -m)"
```

modules/install-consul/install-consul
```diff
@@ -166,8 +166,46 @@ function fetch_binary {
...
+      #
+      # Use Armelv5 for all 32-bit armel systems
+      # Use Armhfv6 for all armhf systems with v6+ architecture
+      # Use Arm64 for all v8 64-bit architectures
+      # The following commands can help determine the right version for your system:
+      #
+      # $ uname -m
+      # $ readelf -a /proc/self/exe | grep -q -c Tag_ABI_VFP_args && echo "armhf" || echo "armel"
+      #
+      local -r vfp_tag="$(readelf -a /proc/self/exe | grep -q -c Tag_ABI_VFP_args)"
```
brikis98, line 194:
```suggestion
      local vfp_tag
      vfp_tag="$(readelf -a /proc/self/exe | grep -q -c Tag_ABI_VFP_args)"
```

modules/install-consul/install-consul
```diff
@@ -166,8 +166,46 @@ function fetch_binary {
...
+      if [[ ! $vfp_tag  ]]; then
```
brikis98, line 195: Are you checking for an empty string here? If so, `-z` is the typical way to do that.

brikis98 (commented):
Thanks for the PR!

modules/install-consul/install-consul
izzyleung: Yep, applied the suggestion and used `-z` for checking empty string. 👍

modules/install-consul/install-consul
izzyleung: Applied the suggestion 🥂

brikis98 (approved):
Great, thank you! I'll kick off tests now.

brikis98:
There was a transient test failure on the first run, but on re-run, all tests passed. Merging now!

brikis98:
https://github.com/hashicorp/terraform-aws-consul/releases/tag/v0.9.3

PR #226, opened 2021-06-01
Migrate to templatefile function from deprecated template_file provider.

lsc:
The provider `template_file` is deprecated and it is encouraged to migrate to the built in `templatefile()` function.

As `template_file` is archived it also means that it's lacking builds for M1 macs.

### DIFF
--- a/examples/example-with-custom-asg-role/main.tf
+++ b/examples/example-with-custom-asg-role/main.tf
@@ -45,8 +45,17 @@ module "consul_servers" {
   cluster_tag_key   = var.cluster_tag_key
   cluster_tag_value = var.cluster_name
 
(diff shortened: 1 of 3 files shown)


brikis98 (commented):
Thanks for the fix! Overall, changes LGTM.

One request: please run `terraform fmt` on all the code, our our pre-commit hook will fail.

lsc:
> Thanks for the fix! Overall, changes LGTM.
> 
> One request: please run `terraform fmt` on all the code, our our pre-commit hook will fail.

Will do!

brikis98:
Ah, OK, great. Kicking off tests now.

brikis98:
Tests passed! Merging now.

PR #204, opened 2021-01-14
Update setup-systemd-resolved

reedlaw:
Select yum or apt-get. Addresses #203

PR #228, opened 2021-06-11
Rebase be39b6d - select yum or apt-get in systemd-resolved

tonyhb:
Rebase of https://github.com/hashicorp/terraform-aws-consul/pull/204 for tests.

### DIFF
--- a/modules/setup-systemd-resolved/setup-systemd-resolved
+++ b/modules/setup-systemd-resolved/setup-systemd-resolved
@@ -67,14 +67,30 @@ function assert_not_empty {
   fi
 }
 
+function has_yum {
+  [ -n "$(command -v yum)" ]
+}
+
+function has_apt_get {
+  [ -n "$(command -v apt-get)" ]
+}
+
 function install_dependencies {
   local -r consul_ip="$1"
 
   log_info "Installing dependencies"
-  sudo apt-get update -y
-  echo iptables-persistent iptables-persistent/autosave_v4 boolean true | sudo debconf-set-selections
-  echo iptables-persistent iptables-persistent/autosave_v6 boolean true | sudo debconf-set-selections
-  sudo apt-get install -y iptables-persistent
+  if has_apt_get; then
+    sudo apt-get update -y
+    echo iptables-persistent iptables-persistent/autosave_v4 boolean true | sudo debconf-set-selections
+    echo iptables-persistent iptables-persistent/autosave_v6 boolean true | sudo debconf-set-selections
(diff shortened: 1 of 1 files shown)


modules/setup-systemd-resolved/setup-systemd-resolved
```diff
@@ -67,14 +67,30 @@ function assert_not_empty {
   fi
 }
 
+function has_yum {
+  [ -n "$(command -v yum)" ]
```
brikis98, line 71: NIT: we follow the [Google Bash style guide](https://google.github.io/styleguide/shellguide.html), so would you mind updating to `[[ ]]` instead of `[ ]`?

modules/setup-systemd-resolved/setup-systemd-resolved
```diff
@@ -67,14 +67,30 @@ function assert_not_empty {
...
 function install_dependencies {
   local -r consul_ip="$1"
 
   log_info "Installing dependencies"
-  sudo apt-get update -y
-  echo iptables-persistent iptables-persistent/autosave_v4 boolean true | sudo debconf-set-selections
-  echo iptables-persistent iptables-persistent/autosave_v6 boolean true | sudo debconf-set-selections
-  sudo apt-get install -y iptables-persistent
+  if has_apt_get; then
+    sudo apt-get update -y
```
brikis98, line 83: NIT: Not related to this PR, but I think we want:

```suggestion
    sudo DEBIAN_FRONTEND=noninteractive apt-get update -y
```

Without it, we may occasionally get an interactive prompt here, causing the build to hang on input.

modules/setup-systemd-resolved/setup-systemd-resolved
```diff
@@ -67,14 +67,30 @@ function assert_not_empty {
...
+    echo iptables-persistent iptables-persistent/autosave_v4 boolean true | sudo debconf-set-selections
+    echo iptables-persistent iptables-persistent/autosave_v6 boolean true | sudo debconf-set-selections
+    sudo apt-get install -y iptables-persistent
```
brikis98, line 86: NIT: Not related to this PR, but I think we want:

```suggestion
    sudo DEBIAN_FRONTEND=noninteractive apt-get install -y iptables-persistent
```

Without it, we may occasionally get an interactive prompt here, causing the build to hang on input.

brikis98 (commented):
Thanks for the PR! Code changes LGTM, except for a couple tiny NITs. Once those are fixed, I can kick off tests.

PR #232, opened 2021-07-10
use `-raw` when getting values from `terraform output`

rileytg:
this prevent the extra `"..."` quotes which messes up the `aws ec2 describe-instances` call


brikis98 (approved):
Thanks for the PR! Changes LGTM. I'll kick off tests.

brikis98:
Tests passed! Merging now, thanks!

